PR #37, opened 2023-08-13
Patch Pickups Attempt 1

dyceron:


input.json
```diff
@@ -11,70 +11,25 @@
...
-        {
-            "actor": {"scenario": "s067_area6c", "actor": "LE_PowerUp_PlasmaBeam"}, 
-            "type": "powerup_gravitysuit"
+            "pickup_actor": {
+                "scenario": "s000_surface",
+                "actor": "LE_PowerUP_Morphball"
+            },
+            "caption": "Plasma Beam acquired.",
+            "model": "powerup_plasmabeam",
+            "item_id": "powerup_plasmabeam",
```
duncathan, line 31: surely this should be `ITEM_WEAPON_PLASMA_BEAM` or something that looks like that

duncathan (changes requested):

input.json
dyceron: oh right thats what item_id is. forgot

open_samus_returns_rando/samus_returns_patcher.py
```diff
@@ -60,37 +62,86 @@ def _wrap(v: str):
...
 
-        model_name: str = pickup["type"]
-        pickup_type = get_data(model_name)
-
         found_actor = False
         for actors in scenario.raw.actors:
             if actor_name in actors:
                 actor = actors[actor_name]
-                actor["type"] = pickup["type"]
+                actor["model"] = pickup["model"]
```
duncathan, line 83: this line should be further down, next to where you edit `actor.type`. also, you may prefer to write it like this instead: `actor.model = pickup["model"]`. (you can't always use these syntaxes interchangeably! but for most things returned by MEDS you can)

open_samus_returns_rando/samus_returns_patcher.py
```diff
@@ -60,37 +62,86 @@ def _wrap(v: str):
...
                 found_actor = True
                 break
     
         if not found_actor:
-            raise KeyError("Actor named '{}' found in ".format(actor_name, actor_reference["scenario"]))
+            raise KeyError("Actor named '{}' found in ".format(actor_name, actor_reference["scenario"]))
```
duncathan, line 88: `raise KeyError(f"No actor named '{actor_name}' found in {actor_reference['scenario']")`

open_samus_returns_rando/samus_returns_patcher.py
duncathan: actually, on second glance - is this line even supposed to be here at all? you might actually want to just remove it altogether

open_samus_returns_rando/samus_returns_patcher.py
```diff
@@ -60,37 +62,86 @@ def _wrap(v: str):
...
     
         if not found_actor:
-            raise KeyError("Actor named '{}' found in ".format(actor_name, actor_reference["scenario"]))
+            raise KeyError("Actor named '{}' found in ".format(actor_name, actor_reference["scenario"]))      
+
+        model_name: str = pickup["model"]
+        model_data = get_data(model_name)
+
+        new_template = copy.deepcopy(template_bmsad)
+        new_template["name"] = f"randomizer_powerup_{i}"
```
duncathan, line 94: remove this line; we'll edit this further down

open_samus_returns_rando/samus_returns_patcher.py
```diff
@@ -60,37 +62,86 @@ def _wrap(v: str):
...
+            item_id = "fMaxEnergy"
+            quantity *= 50.0
+            set_custom_params["Param4"]["value"] = "Full"
+            set_custom_params["Param5"]["value"] = "fEnergy"
+            set_custom_params["Param6"]["value"] = "SPECIALENERGY"
+
+        set_custom_params["Param1"]["value"] = item_id
+        set_custom_params["Param2"]["value"] = quantity
+
+        actordef_id = new_template["name"]
```
duncathan, line 126:
```py
actordef_id = f"randomizer_powerup_{i}"
new_template["name"] = actordef_id
```

duncathan (changes requested):

open_samus_returns_rando/samus_returns_patcher.py
```diff
@@ -60,37 +62,86 @@ def _wrap(v: str):
...
         for actors in scenario.raw.actors:
             if actor_name in actors:
                 actor = actors[actor_name]
-                actor["type"] = pickup["type"]
                 found_actor = True
                 break
     
         if not found_actor:
-            raise KeyError("Actor named '{}' found in ".format(actor_name, actor_reference["scenario"]))
+            raise KeyError(f"No actor named '{actor_name}' found in {actor_reference['scenario']}")
```
duncathan, line 87: actually, this whole block (lines 79-87) can be replaced with this:
```py
actor = next((layer[actor_name] for layer in scenario.raw.actors if actor_name in layer), None)
if actor is None:
    raise KeyError(f"No actor named '{actor_name}' found in {actor_reference['scenario']}")
```

duncathan (changes requested):

duncathan (approved):

PR #52, opened 2023-08-20
Move pickups into pickup.py

dyceron:


ThanatosGit (approved):

open_samus_returns_rando/__init__.py
```diff
@@ -1 +1 @@
-from .samus_returns_patcher import patch
```
ThanatosGit, line 1: I don't think we need this import at all.

dyceron: I just copied from Dread lol

PR #53, opened 2023-08-20
Add ruff

ThanatosGit:


open_samus_returns_rando/__init__.py
```diff
@@ -1 +0,0 @@
-from .samus_returns_patcher import patch_extracted
```
dyceron, line 1: I still don't know about thus @henriquegemignani

dyceron: Dread does this which is obviously more fleshed out https://github.com/randovania/open-dread-rando/blob/main/src/open_dread_rando/__init__.py

dyceron (approved):

PR #56, opened 2023-08-24
Progressive Beams

ThanatosGit:
I copied alot of dread code and refactored it because we need multiple models and multiple items etc.. This is already solved in odr and I don't want to reinvent the wheel.

This heavily breaks the current `input.json` but honestly I'm more interested in getting a working and stable patcher and then integrate it to RDV when schema is stable.


open_samus_returns_rando/pickup.py
```diff
@@ -66,27 +50,110 @@ def patch_pickups(editor: PatcherEditor, pickups_config: list[dict]):
...
+        return bmsad
+
+
+    def patch_multi_item_pickup(self, bmsad: dict) -> dict:
+        pickable: dict = bmsad["components"]["PICKABLE"]
+        script: dict = bmsad["components"]["SCRIPT"]
+
+        set_custom_params: dict = pickable["functions"][0]["params"]
+        # FIXME: I would set it to "ITEM_NONE" but the game is crashing if you pick up a second progressive
+        set_custom_params["Param1"]["value"] = "ITEM_WEAPON_WAVE_BEAM"
```
ThanatosGit, line 67: This is sadly very ugly atm.

The game always opens the inventory / samus screen and show you what you collected with the text on the bottom. If I set it to "ITEM_NONE" it kinda saves a "I want to show you something" state.
The next time you pick another progressive up, it just opens that inventory screen again, adds the new item without highlighting it and just stucks there and doesn't take any input. So, you have to close emulator.

It even persists after save and reload but NOT after save, go to menu and load.

Maybe we could do an ugly hack like giving a missile expansion with an amount 0 or sth. Dunno. Can look at it in the next days.

open_samus_returns_rando/pickup.py
```diff
@@ -1,53 +1,37 @@
...
+        self.pickup = pickup
+        self.pickup_id = pickup_id
+        self.configuration = configuration
 
+    def patch(self, editor: PatcherEditor):
+        raise NotImplementedError
 
-def patch_pickups(editor: PatcherEditor, pickups_config: list[dict]):
-    template_bmsad = editor.get_parsed_asset("actors/items/powerup_chargebeam/charclasses/powerup_chargebeam.bmsad").raw
+# Note: MSR has a lot of customised Lua for Pickups. We maybe need to add them / take the logic into account?
```
dyceron, line 26: If this is talking about the pickups folder, i think we would just remove those and it would be included in `randomizer_powerup.lua`. I only had those files to temporarily test and modify certain pickups.

open_samus_returns_rando/pickup.py
dyceron: I might have a solution for this

dyceron (approved):

PR #59, opened 2023-08-25
Fix Damage Sound if collecting a Varia in heat or Gravity under lava

dyceron:


open_samus_returns_rando/files/randomizer_powerup.lua
```diff
@@ -90,7 +90,8 @@ function RandomizerPowerup.ChangeSuit()
     local model_updater = Game.GetPlayer().MODELUPDATER
     for _, suit in ipairs(suits) do
         if suit.model == model_updater.sModelAlias then break end
-        if Game.GetItemAmount(Game.GetPlayerName(), suit.item) > 0 then
+        if Game.GetItemAmount(Game.GetPlayerName(), suit.item) > (0 or 1) then
```
ThanatosGit, line 93: This is logically the same as before.
`(0 or 1)` returns the first value if it's a truthy value otherwise the second. Numbers are always truthy values in Lua, which means  it always returns 0 which makes it the same check as before
`if Game.GetItemAmount(Game.GetPlayerName(), suit.item) > 0 then`
See the fourth line in the examples: https://www.lua.org/pil/3.3.html

PR #70, opened 2023-08-28
Support option for muting the in game music

dyceron:
Includes Area tracks, Metroid/Boss fights/Samus dying, item fanfares, etc


ThanatosGit (approved):
Metroid without music? What kind of madman would do such a thing?

Has some merge conflicts now because of the other merged pr.

dyceron:
Since game patches now exists, should this go in there instead?

dyceron:
This would make more sense as a cosmetic option. I'm not sure if that would be added to game patches or not.

ThanatosGit:
Could move this and the map thing to a cosmetic option section.

Issue #40, opened 2023-08-14
Suits, Aeion Abilities, and the Baby require a reload to be visible/use


dyceron:

duncathan:
this is probably going to need the same fix(es) as dread did

ThanatosGit:
Is this still the case for Aeion? The others are fixed afaik

dyceron:
It depends on the template bmsad used iirc

dyceron:
The Baby is not as issue as of #72

dyceron:
Suits have been fixed as of #64 and #67

dyceron:
Baby still seems to be broken

ThanatosGit:
A little bit more specific please :p

Afaik all test jsons for the models still work. Is it a specific area, location or state?

dyceron:
Baby on a non custom actor doesn't spawn it in. So far tested on the etank before charge beam in surface.

ThanatosGit:
Could be morph state then? Iirc samus morphed and unmorphed are two different things. Maybe the hatchling component doesn't exist on the morphed variant.

dyceron:
I tested non morphed (with cheats) and it was also tested with morph

ThanatosGit:
Best answer always: Works for me


https://github.com/randovania/open-samus-returns-rando/assets/117127188/4156276c-9427-41e1-9851-033e859fb51c

PR #72, opened 2023-08-30
Support Shuffled Baby

### LINKED ISSUE #18: Allow for the Baby Metroid to be shuffled

dyceron:
The Baby location can now be shuffled, as well at the Baby being shuffle-able to other pickup locations

Fixes #18 
Related to #40

### DIFF
--- a/open_samus_returns_rando/files/levels/s100_area10.lua
+++ b/open_samus_returns_rando/files/levels/s100_area10.lua
@@ -347,6 +347,7 @@ function s100_area10.OnQueenGenerated(_ARG_0_, _ARG_1_)
   end
 end
 function s100_area10.OnQueenDead(_ARG_0_)
+  Scenario.WriteToBlackboard("firstTimeMetroidHatchlingIntroPlayed", "b", true)
   Blackboard.SetProp("DEFEATED_ENEMIES", "Queen", "i", 1)
   Game.SetSubAreaCurrentSetup("collision_camera_020", "PostMetroids_001", true)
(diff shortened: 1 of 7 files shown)


open_samus_returns_rando/files/levels/s100_area10.lua
ThanatosGit, line 1: Could you explain the logic? I can't really follow.
When is `s100_area10.InitFromBlackboard()` called? I assume the first time you enter the area.
The `firstTimeMetroidHatchlingIntroPlayed` should always be true in that case, so why do we need a check for it?
Is `s100_area10.OnMetroidHatchlingIntroCutsceneLaunch()` really called? I thought we just disabled it at all by not commenting out the launch of the cutscene.

dyceron: `s100_area10.InitFromBlackboard()` is called when entering the area, correct. I then check to see if it's true so when you enter `collistion_camera_22` (where the cutscene plays), the egg from which the baby would spawn gets broken. Maybe this can be simplified further but the game checks for the hatchingling cutscene to be played to break the eggs, so this is how I went about it.

ThanatosGit: Yeah, but I'm still wondering if `s100_area10.OnMetroidHatchlingIntroCutsceneLaunch()` is really called because we never call the cutscene.

ThanatosGit (approved):

open_samus_returns_rando/files/levels/s100_area10.lua
dyceron: If it ends up being an issue later we can change this.

Issue #57, opened 2023-08-25
Improve traversal between areas


dyceron:
### Area 1
The Crumble blocks leaving the area forces having Spider or Clipping to leave.

**Solution**: Remove these blocks so only Morph Ball is required.

### Charge Doors
These are at the entrance/exit of every except Area 8 and Surface West. It pretty much forces needing Charge or PBs or Beam Burst early to just travel into areas.

**Solution**: Change these doors into Power Doors

### Grapple Blocks
Leaving Areas 5-7 requires moving a Grapple block to access the next area. This is especially bad for random start/elevators, since backwards traversal in these areas is impossible unless you enable tricks, which is not a good thing IMO.

**Solution**: Set the blocks to post-state (pulled or broken)

-----
Should these be toggleable? Does it enhance the rando to keep/remove them? Having a mostly open path to get around the map could improve the experience a lot.

dyceron:
Charge doors to change:
- Surface: `Door004`, `Door011`
- Area1: `Door002`
- Area2c: `Door001`, `Door007`
- Area3: `Door002`, `Door008`
- Area4: `Door006`, `Door014`
- Area5: `Door006`
- Area6: `Door001`
- Area7: `Door010`
- Area9: `Door012`

PR #76, opened 2023-08-31
Support option to remove grapple blocks

dyceron:
Related to #57

Shortcut option disabled
Metroid Samus Returns_30 08 23_21 05 26 796

Shortcut option enabled
Metroid Samus Returns_30 08 23_21 04 45 363

### DIFF
--- a/open_samus_returns_rando/patcher_editor.py
+++ b/open_samus_returns_rando/patcher_editor.py
@@ -56,3 +56,10 @@ class PatcherEditor(FileTreeEditor):
         newActor.z = coords[2] + offset[2]
 
         return newActor
+    
+    def remove_entity(self, reference: dict):
+        scenario = self.get_scenario(reference["scenario"])
+        layer = reference.get("layer", "default")
+        actor_name = reference["actor"]
+
+        scenario.raw.actors[layer].pop(actor_name)
--- a/open_samus_returns_rando/patcher_editor.py
+++ b/open_samus_returns_rando/patcher_editor.py
@@ -47,7 +47,7 @@ class PatcherEditor(FileTreeEditor):
         self.memory_files = {}
 
     def copy_actor(self, scenario: str, coords, template_actor: Container, new_name: str,
-                   layer_index: int, offset: tuple = (0, 0, 0)):       
+                   layer_index: int, offset: tuple = (0, 0, 0)):
         new_actor = copy.deepcopy(template_actor)
         current_scenario = self.get_scenario(scenario)
         current_scenario.raw.actors[layer_index][new_name] = new_actor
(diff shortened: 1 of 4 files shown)


open_samus_returns_rando/patcher_editor.py
```diff
@@ -56,3 +56,10 @@ def copy_actor(self, scenario: str, coords, templateActor: Container, newName: s
         newActor.z = coords[2] + offset[2]
 
         return newActor
+    
+    def remove_entity(self, reference: dict):
```
ThanatosGit, line 60: The actor should also be removed from the actor group / entity group or whatever you want to call it.
Basically the inverse of where we add them like for the spawn points: https://github.com/randovania/open-samus-returns-rando/blob/b90cf9f1970132d8e1fe1e7cb06c6be6591e7318/open_samus_returns_rando/samus_returns_patcher.py#L89-L91

open_samus_returns_rando/patcher_editor.py
```diff
@@ -56,3 +56,10 @@ def copy_actor(self, scenario: str, coords, templateActor: Container, newName: s
...
+        scenario = self.get_scenario(reference["scenario"])
+        layer = reference.get("layer", "default")
```
ThanatosGit, line 62: Either make sure that layer always exists in the reference or use `0` as default.
The layers in SR have no names, they are indeed based. That's a copy of dread but it doesn't take the difference into account :p

ThanatosGit (changes requested):

open_samus_returns_rando/patcher_editor.py
ThanatosGit: TL;DR: `layer = reference["layer"]`

On a second thought: I would prefer to not use any default here.
If we want to remove an entity, we have to specifiy on which layer it is.

Layer 0 contains the triggers and so it will fail anyways for all non-trigger entities we want to remove, if they haven't specified the layer.

open_samus_returns_rando/patcher_editor.py
ThanatosGit: My new PR should help with that. Could use `remove_actor_from_all_groups`

open_samus_returns_rando/patcher_editor.py
```diff
@@ -47,7 +47,7 @@ def flush_modified_assets(self):
         self.memory_files = {}
 
     def copy_actor(self, scenario: str, coords, template_actor: Container, new_name: str,
-                   layer_index: int, offset: tuple = (0, 0, 0)):
+                   layer_index: int, offset: tuple = (0, 0, 0)):
```
ThanatosGit, line 50: Looks like a weird change. Other changes looks fine to me

ThanatosGit (approved):

PR #82, opened 2023-09-04
Option to make Missile shields not open to Super Missiles

dyceron:
https://github.com/randovania/open-samus-returns-rando/assets/38679103/8c9a8cb6-19c3-4f98-9513-86b11fa6a322


ThanatosGit (approved):
Reminds me that we need to figure how to use supers without missiles because otherwise this option does nothing logically :(

PR #100, opened 2023-09-22
Give all Items from via Lua

ThanatosGit:
It shows the "OK" button at the message box directly but don't know how to change that.


open_samus_returns_rando/files/randomizer_powerup.lua
```diff
@@ -139,192 +126,161 @@ function RandomizerPowerup.IncreaseAmmo(resource)
...
-    if main_item then
-        new_current = current_max + locked_pbs
-    else
-        new_current = current_current + locked_pbs
-    end
-    RandomizerPowerup.SetItemAmount("ITEM_WEAPON_POWER_BOMB_MAX", current_max + locked_pbs)
-    RandomizerPowerup.SetItemAmount("ITEM_WEAPON_POWER_BOMB_CURRENT", new_current)
-    RandomizerPowerup.SetItemAmount("ITEM_RANDO_LOCKED_PBS", 0)
-    hud.UpdatePlayerInfo(true)
+-- TODO: What's the max energy?
```
ThanatosGit, line 129: ?

open_samus_returns_rando/files/randomizer_powerup.lua
```diff
@@ -139,192 +126,161 @@ function RandomizerPowerup.IncreaseAmmo(resource)
...
+    new_max = math.min(new_max, MAX_ENERGY)
+    RandomizerPowerup.SetItemAmount("ITEM_MAX_LIFE", new_max)
+    RandomizerPowerup.SetItemAmount("ITEM_CURRENT_LIFE", new_max)
+
+    local life = Game.GetPlayer().LIFE
+    life.fMaxLife = new_max
+    life.fCurrentLife = new_max
+end
+
+-- TODO: What's the max aeion?
```
ThanatosGit, line 143: ?

open_samus_returns_rando/files/randomizer_powerup.lua
dyceron: 1099

open_samus_returns_rando/files/randomizer_powerup.lua
dyceron: 2200

dyceron:
I assume the max values for energy and aeion aren't hard caps, because I'm able to set my max to be much higher than these.

open_samus_returns_rando/files/randomizer_powerup.lua
ThanatosGit: Does the game do not explode at refill stations? Iirc that was the issue with dread: You could go higher but have crashes at refill stations.

dyceron: I don't recall having issues

ThanatosGit:
Thanks for the fixes.
`ITEM_CURREITEM_CURRENT_SPECIAL_ENERGYNT_LIFE` that happens way too often for me.
Copy paste something from the left to the right, but miss the double click on the right when selecting the string to replace. 😔

dyceron (approved):

PR #102, opened 2023-09-23
Support option for removing crumble blocks in S and A1

dyceron:
Adds an option to change the crumble blocks after Scan Pulse and leaving Area 1 to be Power Beam blocks. This is to open up the rando a bit more. These changes affect the following:
- For Surface, this means that a player can more safely access the Alpha section without being forced to have/find Charge Beam to leave the area.
- For Area 1, this means that Spider Ball is no longer hard required to leave the area. Even though Phase Drift can be an alternative to Spider, Spider is more likely to be placed early due to its overall usefulness.

I couldn't figure out how to delete the blocks and remove the textures, as the textures seem to be left behind. A more ideal situation would be to remove the blocks entirely. Might also be worth splitting this option up in the future, as they are similar, but achieve different goals. This also would affect trick logic, as there are a few methods to bypass the Area 1 blocks with tricks, but this makes most of those pointless, which IMO is not a good thing.

Related to #57

### DIFF
--- a/open_samus_returns_rando/specific_patches/game_patches.py
+++ b/open_samus_returns_rando/specific_patches/game_patches.py
@@ -1,4 +1,4 @@
-from mercury_engine_data_structures.formats import Bmsad
+from mercury_engine_data_structures.formats import Bmsad, Bmsbk
 
 from open_samus_returns_rando.patcher_editor import PatcherEditor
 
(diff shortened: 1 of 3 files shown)


open_samus_returns_rando/specific_patches/game_patches.py
```diff
@@ -62,22 +70,48 @@ def _remove_pb_weaknesses(editor: PatcherEditor):
...
+        post_scan_pulse_crumbles["types"][0]["blocks"][2]["unk4"] = 0.0
+
+    # Crumble blocks leaving Area 1
+    if configuration["patch_area1_crumbles"]:
+        area1 = editor.get_file(
+            "maps/levels/c10_samus/s010_area1/s010_area1.bmsbk", Bmsbk
+        )
+        area1_chozo_seal_crumbles = area1.raw["block_groups"][19]
+        area1_chozo_seal_crumbles["types"][0]["block_type"] = "power_beam"
+        area1_chozo_seal_crumbles["types"][0]["blocks"][0]["unk4"] = 0.0
```
ThanatosGit, line 113: That's the respawn time?
Then we should rename it in MEDS :)

ThanatosGit: Done

ThanatosGit:
Should we merge this?

dyceron:
If it still works with the change then yes. I have nothing else to add at the moment

ThanatosGit (approved):

PR #103, opened 2023-09-23
Add ITEM_MAX_SPECIAL_ENERGY as required

ThanatosGit:
My patcher file was not complaining but the patcher crashed. Looks like it is now required :p


dyceron:
Might this be because we now have the "pop" for aeion?

ThanatosGit:
Yeah, that's why it crashed. I much prefer a validation error (which also tells me my vs code while editing) instead of a random code crash => That's why I added it to the required section in the schema

dyceron (approved):

PR #107, opened 2023-09-25
Door and Shield Patches

ThanatosGit:
Needs to be tested at some point.

I had my favourite bug that a door creature (actor) disappeared when entering a room BUT it was only happening when I glitched through the wall with a cheat. 
Maybe the game is just confused when I get to a room where the only way to this room was via a door, which I'm passing by moving above it through the wall


ThanatosGit:
I think it's Ok for a first version.
I'm pretty sure that there will be fixes later anyways.
(Maybe Squash instead of merging 16 commits?)

dyceron (approved):

PR #113, opened 2023-09-30
Connect Surface East and West until all DNA is collected

### LINKED ISSUE #41: Prevent access to "fake" West Surface from East Surface until Ridley condition is met
Warp player to other Surface scenario on both sides until enough DNA requirement is met. Also solves issue of duplicate items off of the Landing Site

Prevents this https://youtu.be/gOaW8hpZhUc

dyceron:
Warps Samus to the proper room in the opposite Surface scenario to prevent early access to Ridley as part of the DNA goal.
Fixes #41 

https://github.com/randovania/open-samus-returns-rando/assets/38679103/2964c3ab-5b15-4a9d-890e-767aff4f0543


open_samus_returns_rando/files/levels/s000_surface.lua
```diff
@@ -457,6 +457,9 @@ function s000_surface.OnSubAreaChange(_ARG_0_, _ARG_1_, _ARG_2_, _ARG_3_, _ARG_4
   elseif (_ARG_0_ == "collision_camera_024" and not Scenario.ReadFromBlackboard("alpha_killed", false)) then
     s000_surface.LaunchFirstTimeAlphaPresentation()
   end
+  if _ARG_0_ == "collision_camera_000" and _ARG_2_ == "collision_camera_017" then
+    Game.LoadScenario("c10_samus", "s110_surfaceb", "", "Samus", 1)
```
ThanatosGit, line 461: Looks kinda scary to me to not define a start point. I'm even impressed that the game don't just move you to the void or to (0,0,0) coordinates.
If we ever glitch out by this teleport, we probably know why

open_samus_returns_rando/files/levels/s110_surfaceb.lua
```diff
@@ -555,7 +551,11 @@ function s110_surfaceb.OnSubAreaChange(_ARG_0_, _ARG_1_, _ARG_2_, _ARG_3_, _ARG_
   Game.SetScenarioItemEnabledByName("ray01", false)
   Game.SetSceneGroupEnabledByName("sg_debris02", false)
   Game.SetSceneGroupEnabledByName("sg_debris03", false)
-  if _ARG_2_ == "collision_camera_021" and _ARG_0_ == "collision_camera_021" and _ARG_3_ == "RidleyCombat" and Game.GetPlayer().INVENTORY:GetItemAmount("ITEM_METROID_COUNT") == 0 then
+  if _ARG_0_ == "collision_camera_017" and _ARG_2_ == "collision_camera_000" and Blackboard.GetProp("PLAYER_INVENTORY", "ITEM_ADN") < 39 then
```
ThanatosGit, line 554: Do you know when the Blackboard property get's updated? Normally we ask the inventory (which is also the thing we update) by using `RandomizerPowerup.GetItemAmount(item_id)` which is just shorthand for `Game.GetItemAmount(Game.GetPlayerName(), item_id)`

Might be synced by the game at the same time but I don't know.

open_samus_returns_rando/files/levels/s110_surfaceb.lua
```diff
@@ -555,7 +551,11 @@ function s110_surfaceb.OnSubAreaChange(_ARG_0_, _ARG_1_, _ARG_2_, _ARG_3_, _ARG_
...
+    Game.LoadScenario("c10_samus", "s000_surface", "", "Samus", 1)
+    Game.GetPlayer().vPos = V3D(-22800, 4450, 0)
```
ThanatosGit, line 556: ...see previous comment: I guess you are in the void without setting the vPos?

open_samus_returns_rando/files/levels/s000_surface.lua
dyceron: So what's happens is it warps to the scenario, and then since it doesn't have an actor listed, it just loads samus at coordinates where the load happened.  Since both Surfaces line up there, it's a perfect match. Going back into East I set coordinates because otherwise Samus warp position is inside the baby blocks, so this moves her outside the blocks barely.

open_samus_returns_rando/files/levels/s110_surfaceb.lua
dyceron: I'm not sure when it updates, but this was the only way I could think of to get this to work. Could easily test this by starting with 38 dna and then picking up a 39th dna in the section before this and see how it behaves

ThanatosGit: I really prefer to be consistent with the usage. Makes it more readable. :p
But I guess the blackboard property get's updated at the same time.

Wouldn't the easiest way have been to just replace `ITEM_METROID_COUNT` with the `ITEM_ADN`? The line already contained a way on how to obtain the amount of an item.

dyceron: I tried variations of that and also checking for other items and no matter what I did, the condition would never be met and thus never initiate that function. I do not understand why. This was the only method that worked for me.

ThanatosGit: Worked for me with `Game.GetItemAmount(Game.GetPlayerName(), "ITEM_ADN")`

ThanatosGit (approved):

PR #115, opened 2023-09-30
Remove new doors from entity groups

ThanatosGit:
It sucks but I run into some issues again. Wtf is wrong with adding actors to entity groups  😥

(two lines were too long according to ruff)

### DIFF
--- a/open_samus_returns_rando/specific_patches/door_patches.py
+++ b/open_samus_returns_rando/specific_patches/door_patches.py
@@ -107,13 +107,15 @@ def _patch_beam_covers(editor: PatcherEditor):
         "s000_surface": ["LE_PlasmaShield_Door_008"],
         "s010_area1": ["LE_DoorShieldWave008", "LE_SpazerShield_Door007"],
         "s020_area2": ["LE_WaveShield_Door010"],
-        "s025_area2b": ["LE_PlasmaShield_Door017", "LE_WaveShield_Door008", "LE_WaveShield_Door016", "LE_WaveShield_Door019"],
+        "s025_area2b": ["LE_PlasmaShield_Door017", "LE_WaveShield_Door008",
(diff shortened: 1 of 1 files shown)


dyceron:
What's the downside of not having these in groups? I'm also curious which doors are giving you issues.

ThanatosGit:
I know that the game only activates entites which are part of the current active collision camera plus/minus some extra things if you just left a room / collision camera.

Adding them to no entity group makes them global and always active. It basically will consume unecessary resources.
My knowledge about engines is like zero. It maybe goes through all kind of animation loops, keeps the model in memory, keeps the hitbox active etc. but probably will not be rendered unless it is in the viewporrt.
If a beam had this range (and a player the accuracy :p) , you could maybe shoot the door from like 500 rooms away?

> I'm also curious which doors are giving you issues.

My tests in the last days took only place in 4 rooms. And 2 out of 2 doors had issues. The metroid, with the short introduction out of the water, in my metroid test vidoes, only spawns if the door at the other metroid is destroyed or if you get to the room from the right but not from the left.
The door in surface, where you put the transition to s110, can be destroyed but as soon as you enter all actors are removed in the room with the pickup and also all actors in the room you were coming from when leaving.

dyceron:
> The metroid, with the short introduction out of the water, in my metroid test vidoes, only spawns if the door at the other metroid is destroyed or if you get to the room from the right but not from the left.
Not sure which one this is. 
> The door in surface, where you put the transition to s110, can be destroyed but as soon as you enter all actors are removed in the room with the pickup and also all actors in the room you were coming from when leaving.
Honestly for this one I'd just exclude it. It's a small puzzle room and I highly doubt we'll spawn in here anyway. Especially not from s100, only in s000. Could just keep the double for s000.

dyceron (approved):
I guess this will do for now. I suppose this will have to be revisited later for when we add Door Lock Rando. Hopefully not though.

PR #119, opened 2023-10-01
Hard Code Baby Pickup

dyceron:


ThanatosGit (commented):
Hmmm I feel the custom_pickups could still come in handy for debugging.
As long as they are not required in the schema, it won't hurt to keep them.
Especially, considering that we could remove `bmsad` file for the model dependencies and also some stuff, we maybe just added because of the old pkg bug.

dyceron:
That's fair. Only annoying thing is it's extra code in the patcher file that isnt used unless we debug then.

ThanatosGit:
If `custom_pickups` is an empty array, it will just skip the loop because there are no items. Totally fine to keep it in.

dyceron:
I meant the code that I removed from `samus_returns_patcher`

open_samus_returns_rando/custom_pickups.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def _get_actor(editor: PatcherEditor):
+    actor_to_copy = {
+        "scenario": "s000_surface",
+        "layer": "9",
+        "actor": "LE_PowerUP_Morphball",
+    }
+    new_actor = editor.resolve_actor_reference(actor_to_copy)
+    return new_actor
```
ThanatosGit, line 11: I just realised that the naming is confusing. This is not a new actor, this is just the actor you use as a reference.

open_samus_returns_rando/custom_pickups.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def _create_baby_pickup(editor: PatcherEditor):
+    name_of_scenario = "s100_area10"
+    scenario_10 = editor.get_scenario(name_of_scenario)
+
+    baby_actor = _get_actor(editor)
```
ThanatosGit, line 18: This is also not the baby_actor but the actor you use as reference.

open_samus_returns_rando/custom_pickups.py
```diff
@@ -0,0 +1,30 @@
...
+    editor.copy_actor(
+        name_of_scenario, (-3400.0, 11225.0, 0.0), baby_actor, "LE_Baby_Hatchling", 9
+    )
```
ThanatosGit, line 21: This is what return the new actor (but as we don't need it anymore in this function it is not assigned to a variable)

ThanatosGit (changes requested):

open_samus_returns_rando/custom_pickups.py
dyceron: What do you mean we don't need it anymore?

ThanatosGit: We don't need the return of it like
```
    baby_actor = editor.copy_actor(
        name_of_scenario, (-3400.0, 11225.0, 0.0), baby_actor, "LE_Baby_Hatchling", 9
    )
```
because we are not doing anything with `baby_actor` after it is created in this function.

dyceron: I'm still not understanding what you want me to fix since this works as it should.

ThanatosGit: Nothing, this was a comment about what actually is the baby_actor :p

dyceron:
Do you want me to readd the function to the patcher file or maybe even to the new file so we can still use this to debug locations? Also anything else I need to change?

ThanatosGit:
Yes, I would prefer to keep it. It has just like 10 LOC and a little bit schema json.
Maybe move it to `specific_patches`?

dyceron:
I could just put it in custom pickups and then call it into the patcher file

dyceron:
Or I'll make a debug file that's for testing, since custom spawn points will be the same thing eventually

ThanatosGit (approved):

PR #120, opened 2023-10-01
Readd starting inventory

dyceron:


open_samus_returns_rando/samus_returns_patcher.py
```diff
@@ -48,9 +48,15 @@ def create_custom_init(configuration: dict) -> str:
...
     inventory_update = {
         "ITEM_MAX_LIFE": max_life,
         "ITEM_MAX_SPECIAL_ENERGY": max_aeion,
+        "ITEM_WEAPON_MISSILE_MAX": 0,
+        "ITEM_WEAPON_SUPER_MISSILE_MAX": 0,
+        "ITEM_WEAPON_POWER_BOMB_MAX": 0,
+        "ITEM_METROID_COUNT": 0,
+        "ITEM_METROID_TOTAL_COUNT": 40,
     }
     inventory.update(inventory_update)
```
ThanatosGit, line 61: This will not work. If your inventory already contains e.g. `"ITEM_WEAPON_MISSILE_MAX"`, you will overwrite it with 0. You disable starting with any of these items by that.

ThanatosGit (changes requested):

open_samus_returns_rando/samus_returns_patcher.py
dyceron: How does Dread bypass this?

ThanatosGit: By doing it the other way around. You update the "must-have" with the starting inventory from the patcher. That way you don't overwrite the `"ITEM_WEAPON_MISSILE_MAX": 0` (so keep it at 0) or it get's overwritten with the value from the patcher.json inventory.
I think, it was at first this way? Afaik I changed it because it wouldn't allow me to define the `ITEM_MAX_LIFE`...but I'm fine with reverting it. So, your starting max life is defined by energy_per_tank + the amount of tanks you start with.

dyceron: Revert as in, go back to what we had before the other pr?

ThanatosGit: Whatever is easier for you.

ThanatosGit (changes requested):
Well, we really need to agree on what is the goal of this PR. 

What must contain the schema, what can be omitted or what is not allowed at all?

With the following lines, it will be still required to have `ITEM_MAX_LIFE` and `ITEM_MAX_SPECIAL_ENERGY` in the patcher.json, which was my intention with the PR that changed the whole inventory part.

If that should not be the case, it must also be removed from the required field in the schema, otherwise the validation will fail.

https://github.com/randovania/open-samus-returns-rando/blob/0f0308d76d7baf9abaabda5a909e0faa0d04ba66/open_samus_returns_rando/samus_returns_patcher.py#L35-L39

dyceron:
Isn't the whole point of having the inventory update to fill in what is missing from RDV?

ThanatosGit:
> Isn't the whole point of having the inventory update to fill in what is missing from RDV?

Yes, but that's why we need an agreement. The current schema requires RDV to export a patcher.json with `ITEM_MAX_LIFE` because it just pops it out of it.
You can have a field in randovania like "Starting energy" where you can type a number.
Or you just completly omit it and always start with 99 energy + the amount of starting e tanks. Which e.g. is what dread does but we are not forced to do it the same way.

dyceron:
> Which e.g. is what dread does but we are not forced to do it the same way.

Dread does have a field for starting with a certain amount of energy though. It's possible I didn't account for everything in `patch_data_factory` regarding that but at the same time I don't think I missed anything for that.

ThanatosGit:
As you know, I made some stuff for RDV. I created a branch here: https://github.com/randovania/randovania/compare/main...ThanatosGit:randovania:msr-things
It ensures that energy and aeion is always present.

For this PR, only change should be to edit the required fields to only contain both.
https://github.com/randovania/open-samus-returns-rando/blob/596b549426fd5e2f41cf90eb5b9ed5975fe865c0/open_samus_returns_rando/files/schema.json#L13-L21

to
```
 "required": [ 
     "ITEM_MAX_LIFE", 
     "ITEM_MAX_SPECIAL_ENERGY",
 ] 
```

dyceron:
What about Metroid counts?

ThanatosGit:
> What about Metroid counts?

You already took care of that with the changes in this pr. It's in the final_inventory variable and therefore not required in the schema / patcher.json

ThanatosGit (approved):

PR #123, opened 2023-10-02
Map ITEM_RANDO_DNA to ITEM_ADN

ThanatosGit:


open_samus_returns_rando/files/schema.json
```diff
@@ -335,7 +335,46 @@
...
+                "ITEM_RANDO_DNA_30",
+                "ITEM_RANDO_DNA_31",
+                "ITEM_RANDO_DNA_32",
+                "ITEM_RANDO_DNA_33",
+                "ITEM_RANDO_DNA_34",
+                "ITEM_RANDO_DNA_35",
+                "ITEM_RANDO_DNA_36",
+                "ITEM_RANDO_DNA_37",
+                "ITEM_RANDO_DNA_38",
+                "ITEM_RANDO_DNA_39"
```
dyceron, line 377: What happens if you don't start with any dna or only some of it? Will the schema throw an error?

ThanatosGit: No, it will not throw an error. This is just the enumeration of what items exists but none of these are required.

dyceron: Ah right.

dyceron (approved):

PR #125, opened 2023-10-02
Random fixes

ThanatosGit:
- Use `TG_Heat_033` because `TG_Heat_004` removed the metroid (this is so stupid).
- Replace `\n` in caption because otherwise it is breaking the lua file.

### DIFF
--- a/open_samus_returns_rando/lua_editor.py
+++ b/open_samus_returns_rando/lua_editor.py
@@ -92,7 +92,7 @@ class LuaEditor:
             "name": class_name,
             "resources": resources,
             "parent": parent,
-            "caption": lua_util.wrap_string(pickup["caption"]),
+            "caption": lua_util.wrap_string(pickup["caption"].replace("\n", "\\n")),
             "sound": lua_util.wrap_string(sound),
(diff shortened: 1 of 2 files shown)


open_samus_returns_rando/lua_editor.py
```diff
@@ -92,7 +92,7 @@ def get_script_class(self, pickup: dict, actordef_name: str = "") -> str:
             "name": class_name,
             "resources": resources,
             "parent": parent,
-            "caption": lua_util.wrap_string(pickup["caption"]),
+            "caption": lua_util.wrap_string(pickup["caption"].replace("\n", " ")),
```
dyceron, line 95: Does this mean the cptions will all be one line and weird?

ThanatosGit: Does the game support multiline captions? 🙃
If yes, I maybe should escape the \n instead of removing it

ThanatosGit: Alright. Adressed

dyceron (approved):

PR #129, opened 2023-10-04
Move baby to samus

### LINKED ISSUE #40: Suits, Aeion Abilities, and the Baby require a reload to be visible/use

ThanatosGit:
Untested. Won't test this today anymore.
At least it works delayed, hopefully it also works directly without delay.

Fixes #40


dyceron (approved):
Quick test shows this works. Baby spawns during the textbox, but that's not a huge deal.

PR #137, opened 2023-10-07
Add missing itemsphere dependency

### LINKED ISSUE #132: Itemsphere dependency issue in area2b
image

dyceron:
This should be the last model dependency PR we need to make 😛 
Fixes #132 

image


ThanatosGit (approved):
Not the last :p.
At some day we should remove all the bmsad stuff as dependency. Also I bet there are other dependencies, which can be removed because they were just added because of the pkg bug with uneven/even added amount of models

PR #139, opened 2023-10-09
Add offset to models

ThanatosGit:
1. I moved the model related logic from `patch_single_item_pickup` to `patch_model` which will fix issues with exporting from randovania.
2. The basic idea (as most items are tank items) is to raise all models a few pixels to match the tanks.

Not perfect but enough for me 😛
image


dyceron:
What happened to scan pulse?

dyceron (approved):

PR #141, opened 2023-10-10
Patch for main items and non required mains

ThanatosGit:
Quick notes:
- `patch_single_item_pickup` and `patch_multi_item_pickup` had no difference anymore (everything handled by the lua and model part of the patcher)
- changed to check for the models instead of the item id in model patching. This is because non required mains need the first item id to be the main item but they are using the model of the tank item (and it is really dependent on the model not the item id)
- i used the plural `_TANKS`  for the names to be in sync with the energy and aeion from vanilla
- had to do a change in the naming scheme for the lua functions, otherwise the lua was using the same function for the main item and tanks with non required mains.


ThanatosGit:
I seriously have no idea what these options are.
Random models sounds like a pretty crappy option. But you missed the `in model patching` part. Especially random models have no association to the item id. The offset and what visual effects should be added are dependent on the model and not the item id.

dyceron:
I imagine if it's using model instead then random models should work in theory

ThanatosGit:
> Was referring to these: grafik
> 
> but yeah, missed the second part

I may or may not have ever looked at any of the shown options 😅

PR #143, opened 2023-10-12
Fix Metroids

### LINKED ISSUE #135: Certain Metroids do not get disabled/drop an item
Sometimes the Metroid function is called and sometimes its not. Problematic Metroids found in testing:

- area2b: Gamma_001
- area6: Gamma_002

dyceron:
- Ensures the Gamma in area2b always spawns regardless of entrance
- Ensures the Omega in area7 always spawns regardless of entrance
- Fixes Metroid radars not ending
- Fixes #135

### DIFF
--- a/open_samus_returns_rando/files/custom_scenario.lua
+++ b/open_samus_returns_rando/files/custom_scenario.lua
@@ -72,8 +72,9 @@ function Scenario.OnSubAreaChange(old_subarea, old_actorgroup, new_subarea, new_
   Scenario.UpdateProgressiveItemModels()
 end
 
-function Scenario.SetMetroidSpawngroupOnCurrentScenario(created_actor, group_name)
+function Scenario.SetMetroidSpawngroupOnCurrentScenario(created_actor, group_name, is_multi)
   if created_actor ~= nil and created_actor.sName ~= nil then
     CurrentScenario.currentMetroidSpawngroup = group_name
+    CurrentScenario.isMultiGamma = is_multi or false
   end
 end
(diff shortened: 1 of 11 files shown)


ThanatosGit (approved):

ThanatosGit:
Works for me with my additional commits. Feel free to confirm with the provided input json

dyceron:
The radar keeps going even after the multi-room Gamma is dead and has given the item. Might be wise to fix it in this PR. Just have to find out why it doesn't stop.

ThanatosGit:
Pushed something for the issue. Didn't test it

dyceron:
Testing the new changes and the radar still plays. Reloading the cc (entering a fadeout) does load the other spawn group but yeah this still isn't disabled.

ThanatosGit:
It is really hard to follow for me with all the name clashes 😔. In the game's wording, the sub area's in a scenario file are just a list of all the cc's with their actors. Which makes it basically: sub area = cc = sometimes we call it room but in randovania wording area.

So... what is "reloading the entire sub area" ? It looks like it means reloading the scenario? Or is really just re-enter a room with a fade out? But that would be no difference to other metroids? Killing them leaves the room empty, reload it via enter/exit another room + walk a few steps away in the other room and then it spawns the normal enemies in the metroid room.

dyceron:
Sorry let me clarify. After defeating the multi-room Gammas, the radar still continues. Reloading the room via a fade out sets the post-gamma spawngroup to be active, but the radar doesn't stop. Reloading the scenario also sets the post-gamma spawngroup to be active, and also disables the radar.

PR #145, opened 2023-10-13
Fix Heatzone damagesound playing with Gravity and not Varia

dyceron:


ThanatosGit (approved):
Funny PR. This should actually never be logically possible as the only supprted way should be progressive suits.
Guess that happens in my test file because I was too lazy to write one more line into the starting inventory section.

PR #150, opened 2023-10-15
Adjust position/rotation for certain pickup actors

dyceron:
- Adjusts the y-position for the Aeion abilities and Spider Ball to be properly aligned.
- Fixes the rotation of Grapple Beam and Space Jump to be `0, 0, 0` from `0, 180, 0`
  - Spring Ball doesn't seem to be affected by this 😞 

image
image
image
image
image


ThanatosGit:
Is there a specific reason why the rotation is changed during patch time but the height during run time? Looks very inconsistent

dyceron:
Height was very easy this way. I couldn't figure out how to do the same for Rotation so I hard-coded those.

ThanatosGit (approved):

PR #155, opened 2023-10-16
Try to fix actors

ThanatosGit:
Oh boy, I have no idea why I did the second commit by "reverting" the changes from the first commit by hand.
It will be handled all by the MEDS functions, so this is basically just a small refactor + activating adding the shields to the entity groups.

P.S. Please squash merge to get rid of the "WIP" commit


dyceron:
Should I wait for the MEDS PR to go through to merge?

dyceron (approved):

ThanatosGit:
Yeah, otherwise we need to start to use my fork of meds again.

PR #156, opened 2023-10-16
Set ammo current value equal to max value on pickup

dyceron:
- Fixes Missiles not properly increasing current ammo when Missile tanks are collected
- Adds QoL where collecting a tank now updates the current value to match the max value, so low ammo counts are less of an issue

### DIFF
--- a/open_samus_returns_rando/files/randomizer_powerup.lua
+++ b/open_samus_returns_rando/files/randomizer_powerup.lua
@@ -41,12 +41,12 @@ function RandomizerPowerup.IncreaseItemAmount(item_id, quantity, capacity)
         local specialEnergy = Game.GetPlayer().SPECIALENERGY
         specialEnergy.fMaxEnergy = capacity
         specialEnergy.fEnergy = capacity
-    elseif item_id == "ITEM_WEAPON_MISSILE_MAX" then
-        itemAmount:SetItemAmount("ITEM_WEAPON_MISSILE_CURRENT", item_id)
-    elseif item_id == "ITEM_WEAPON_SUPER_MISSILE_MAX" then
-        itemAmount:SetItemAmount("ITEM_WEAPON_SUPER_MISSILE_CURRENT", item_id)
-    elseif item_id == "ITEM_WEAPON_POWER_BOMB_MAX" then
-        itemAmount:SetItemAmount("ITEM_WEAPON_POWER_BOMB_CURRENT", item_id)
+    elseif item_id == "ITEM_WEAPON_MISSILE_CURRENT" then
+        itemAmount:SetItemAmount(item_id, capacity)
+    elseif item_id == "ITEM_WEAPON_SUPER_MISSILE_CURRENT" then
+        itemAmount:SetItemAmount(item_id, capacity)
+    elseif item_id == "ITEM_WEAPON_POWER_BOMB_CURRENT" then
+        itemAmount:SetItemAmount(item_id, capacity)
     end
 end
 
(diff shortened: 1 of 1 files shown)


open_samus_returns_rando/files/randomizer_powerup.lua
```diff
@@ -36,10 +36,17 @@ function RandomizerPowerup.IncreaseItemAmount(item_id, quantity, capacity)
...
     RandomizerPowerup.SetItemAmount(item_id, target)
 
-    if  item_id == "ITEM_CURRENT_SPECIAL_ENERGY" then
+    local itemAmount = Game.GetPlayer().INVENTORY
+    if item_id == "ITEM_CURRENT_SPECIAL_ENERGY" then
         local specialEnergy = Game.GetPlayer().SPECIALENERGY
         specialEnergy.fMaxEnergy = capacity
         specialEnergy.fEnergy = capacity
+    elseif item_id == "ITEM_WEAPON_MISSILE_MAX" then
+        itemAmount:SetItemAmount("ITEM_WEAPON_MISSILE_CURRENT", item_id)
```
ThanatosGit, line 45: Is there an example usage of this function like this?
Otherwise I'm pretty sure this works just by accident.

The usage of the function in the lua from the game itself looks like this:
```lua
      Game.GetPlayer().INVENTORY:SetItemAmount("ITEM_WEAPON_MISSILE_MAX", 265, true)
      Game.GetPlayer().INVENTORY:SetItemAmount("ITEM_WEAPON_MISSILE_CURRENT", 265, true)
      Game.GetPlayer().INVENTORY:SetItemAmount("ITEM_WEAPON_SUPER_MISSILE_MAX", 35, true)
      Game.GetPlayer().INVENTORY:SetItemAmount("ITEM_WEAPON_SUPER_MISSILE_CURRENT", 35, true)
      Game.GetPlayer().INVENTORY:SetItemAmount("ITEM_WEAPON_POWER_BOMB_MAX", 20, true)
      Game.GetPlayer().INVENTORY:SetItemAmount("ITEM_WEAPON_POWER_BOMB_CURRENT", 20, true)
      Game.GetPlayer().INVENTORY:SetItemAmount("ITEM_MORPH_BALL", 1, true)
```

but you are using it like `itemAmount:SetItemAmount("ITEM_WEAPON_MISSILE_CURRENT", "ITEM_WEAPON_MISSILE_MAX")`

You are using a string  where a number should be used. Interpreting the string as number will be just a high ass number, so you could also replace it by `99999`.
In fact, I'm kinda scared that using a wrong and much larger data type will explode or lead to undefined behaviour at some other point (but I'm also not an expert in lua integration, which means I have no idea how the library handles such cases)

ThanatosGit (changes requested):

open_samus_returns_rando/files/randomizer_powerup.lua
dyceron: In my short testing this worked just fine, but maybe it isn't ideal. What would be a better way to set the value to match the value of max?

ThanatosGit: Our `RandomizerPowerup.OnPickedUp` first increase the `_MAX` item and then it calls `RandomizerPowerup.IncreaseAmmo`.
In `IncreaseAmmo` it calls `RandomizerPowerup.IncreaseItemAmount(current_id, resource.quantity, resource.item_id)`.

To stay in the same logic pattern, it should increase the current value if it get's called with the `_CURRENT` thing as `item_id` and not with the `_MAX` thing. See how it is done for `ITEM_CURRENT_SPECIAL_ENERGY`. Then it would be (similiar for the other items):
```lua
    elseif item_id == "ITEM_WEAPON_MISSILE_CURRENT" then
        itemAmount:SetItemAmount(item_id, capacity)
```

ThanatosGit: BTW: Does this also works with the special energy? Then everything follows the same pattern and looks nicer :D

ThanatosGit (approved):

PR #157, opened 2023-10-17
Create new checkpoint for entering 2b Gamma from right

dyceron:
Creates a new trigger and two new checkpoints on the right side entrance for the area2b Gamma

### DIFF
--- a/open_samus_returns_rando/specific_patches/metroid_patches.py
+++ b/open_samus_returns_rando/specific_patches/metroid_patches.py
@@ -9,7 +9,7 @@ from open_samus_returns_rando.patcher_editor import PatcherEditor
 
 LOG = logging.getLogger("metroid_patches")
 
-def patch_metroids(editor: PatcherEditor):
+def _patch_metroids(editor: PatcherEditor):
     METROID_FILES = [
         "actors/characters/alpha/charclasses/alpha.bmsad",
         "actors/characters/alphaevolved/charclasses/alphaevolved.bmsad",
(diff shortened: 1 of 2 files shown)


open_samus_returns_rando/specific_patches/metroid_patches.py
```diff
@@ -63,3 +63,59 @@ def patch_metroids(editor: PatcherEditor):
...
+    trigger = _get_trigger(editor)
+    spawnpoint = _get_spawnpoint(editor)
+
+    editor.copy_actor(
+        name_of_scenario, (-1650.0, -4500.0, 0.0), trigger, name_of_trigger, 0,
+    )
+
+    editor.copy_actor(
+        name_of_scenario, (-1500.0, -4500.0, 0.0), spawnpoint, name_of_spawnpoint, 5,
+    )
```
ThanatosGit, line 104: do `spawnpoint_actor = editor.copy_acotr(...)` then you don't need to get the actor again line 116.

open_samus_returns_rando/specific_patches/metroid_patches.py
```diff
@@ -63,3 +63,59 @@ def patch_metroids(editor: PatcherEditor):
...
+
+    editor.copy_actor(
+        name_of_scenario, (-1500.0, -4500.0, 0.0), spawnpoint, name_of_spawnpoint_out, 5,
+    )
+
+    scenario_2b.add_actor_to_entity_groups("collision_camera039", name_of_trigger)
+    scenario_2b.add_actor_to_entity_groups("collision_camera039", name_of_spawnpoint)
```
ThanatosGit, line 111: Oh, did you copy paste that from me? I think it is missing a `_`? Afaik it should be `collision_camera_039`.
We should check if I messed that up in my PR too and maybe also throw an error from MEDS, if it can not find any group at all using this function?

ThanatosGit (changes requested):

open_samus_returns_rando/specific_patches/metroid_patches.py
ThanatosGit: NVM: I already have such a check in MEDS, phew. They really have a different naming scheme in scenario 2b oO.

dyceron:
Next pr that adds checkpoints for other metroids will move the Gamma spawn stuff to the metroid file as well.

PR #158, opened 2023-10-17
Chozo seals

ThanatosGit:
Just opened it to link the issues


ThanatosGit:
Ok, let's start merging the stuff.
This patches the chozo seals to not drain DNA and behave like a save station...but we patched out the save behaviour and show hints.
This is working on the patcher side. Providing a useful hint is up to randovania.

open_samus_returns_rando/files/levels/s000_surface.lua
```diff
@@ -69,6 +69,7 @@ function s000_surface.OnShipStartPointTeleport(_ARG_0_, _ARG_1_)
   end
 end
 function s000_surface.InitFromBlackboard()
+  Scenario.InitGUI()
```
dyceron, line 72: Why does surface not have this placed under main()?

ThanatosGit: It was how I did it initially but changed it and forgot it here.

PR #161, opened 2023-10-18
Add missing checkpoints for most multi-entrance Metroids

dyceron:
Did not test most of these since it requires https://github.com/randovania/mercury-engine-data-structures/pull/88 to be merged first, otherwise rooms break
 
- Added checkpoints to all but two multi-room Metroids (skipped ones connect to an item room so checkpoint isn't that important)
- Renamed Gamma 2b checkpoints to match the vanilla naming scheme


ThanatosGit (approved):
Looks like most, but not all, methods follow the same pattern. Could maybe generalize it later.

PR #162, opened 2023-10-18
Patch functions for RDV export

ThanatosGit:


dyceron (commented):
With the addition of logger.py, should the other files that use logging be updated accordingly?

ThanatosGit:
Oh, good point. Don't know: Do we need to distinguish between logs from all the different files or is it enough for us to just log "samus_returns_patcher"?
I think the latter is fine.

ThanatosGit:
Better to do in another pr as we don't know how to export for 3DS atm.

PR #165, opened 2023-10-19
Deal with weird room being present in surface and surfaceb

ThanatosGit:
- Removed the shield copy
- Ensures that surface and surfaceb is the same item by special case it (which what it really is!)


dyceron (changes requested):
This covers getting either pickup first?

Also you are missing the other Landing Site item here

dyceron (approved):

PR #166, opened 2023-10-21
Remove setup.cfg, update MEDS, add Hooks

ThanatosGit:


dyceron:
Do we not need setup.cfg anymore? What was it used for?

dyceron (approved):

PR #169, opened 2023-10-27
Update Map icons

### LINKED ISSUE #15: Item icon on map does not update when picked up
If a Major Item is shuffled to a tank location, the map icon does not update to indicate it has been picked up.

dyceron:
- Adds new map icon for Baby Metroid pickup
- Changes all item icons to open circles (excluding Aeion Ability Artifacts and Metroids)
- Changes all removed acid tiles to be blue from purple
- Updates door icons to match patcher changes
- Fixes #15 

image

### DIFF
--- a/open_samus_returns_rando/specific_patches/map_icons.py
+++ b/open_samus_returns_rando/specific_patches/map_icons.py
@@ -1,11 +1,24 @@
-from construct import ListContainer
 from mercury_engine_data_structures.formats import Bmsmsd
 
 from open_samus_returns_rando.constants import ALL_AREAS
 from open_samus_returns_rando.patcher_editor import PatcherEditor
 
 
-def _patch_tiles(tiles: ListContainer, editor: PatcherEditor):
+def patch_tiles(editor: PatcherEditor):
+    DOORS = {
+        "s000_surface": ["Door004", "Door011"],
+        "s010_area1": ["Door002", "Door004", "Door012", "Door016"],
+        "s025_area2b": ["Door008"],
+        "s028_area2c": ["Door001", "Door007"],
+        "s030_area3": ["Door002", "Door005", "Door006", "Door008"],
+        "s040_area4": ["Door001", "Door006", "Door014"],
+        "s050_area5": ["Door006"],
+        "s060_area6": ["Door001"],
+        "s067_area6c": ["Door005", "Door006", "Door009"],
+        "s070_area7": ["Door010"],
+        "s090_area9": ["Door012"],
+    }
+
     for area in ALL_AREAS:
         if area != "s110_surfaceb":
             scenario = editor.get_file(f"gui/minimaps/c10_samus/{area}.bmsmsd", Bmsmsd)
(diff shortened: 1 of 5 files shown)


open_samus_returns_rando/specific_patches/map_icons.py
```diff
@@ -0,0 +1,332 @@
...
+    for tile in range(384):
+        # Hazardous Tiles
+        if mapicon[tile]["tile_type"] == "ACID":
+            mapicon[tile]["tile_type"] = "NORMAL"
+        # Items
+        if mapicon[tile]["icons"] != ([]) and mapicon[tile]["icons"][0][
+            "icon"
+        ].startswith("item"):
+            mapicon[tile]["icons"][0]["icon"] = "itemenabled"
+    mapicon[158]["icons"][0]["icon"] = "itemenabledheat"
```
ThanatosGit, line 59: I'm wondering: If the `tile_type` is already heat, what is the difference between `itemenabled` and `itemenabledheat`?

ThanatosGit: Yeah, but what is the difference? Do they have a different color or why are these two different icons? Is the open circle not transparent that it would look blue within the red heat tile?

dyceron: The former is blue and the latter is red. I specified heat afterwards because it first sets to blue which is wrong when on the heat tile.

open_samus_returns_rando/specific_patches/map_icons.py
```diff
@@ -0,0 +1,332 @@
...
+
+    for tile in range(351):
+        # Hazardous Tiles
+        if mapicon[tile]["tile_type"] in {"ACID", "ACID_RISE", "ACID_FALL"}:
+            mapicon[tile]["tile_type"] = "NORMAL"
+        # Items
+        if mapicon[tile]["icons"] != ([]) and mapicon[tile]["icons"][0][
+            "icon"
+        ].startswith("item"):
+            mapicon[tile]["icons"][0]["icon"] = "itemenabled"
```
ThanatosGit, line 281: Is it possible to make the code less verbose?
Some check for `ACID_RISE` and `ACID_FALL`, some not, but can you just do it for all scenarios? Is the `range` parameter just the length?
I'm thinking about one function which get's called for every bmsmsd. If I combined it correctly it would look like this?
```python
def bla(tiles: ListContainer):
    for tile_idx in range(len(tiles)):
        current_tile = tiles[tile_idx]
        # Hazardous Tiles
        if current_tile["tile_type"] in {"ACID", "ACID_RISE", "ACID_FALL"}:
            current_tile["tile_type"] = "NORMAL"

        # Items
        icons = current_tile["icons"]
        if len(icons) != 0 and (
            icons[0]["icon"].startswith("item")
            or icons[0]["icon"].startswith("spenergy")
            or icons[0]["icon"].startswith("itemsphere")
            or icons[0]["icon"].endswith("tank")
        ):
            icons[0]["icon"] = "itemenabled"
```

ThanatosGit (changes requested):

open_samus_returns_rando/specific_patches/map_icons.py
dyceron: Yes the range is the length

open_samus_returns_rando/specific_patches/map_icons.py
```diff
@@ -0,0 +1,172 @@
+from construct import ListContainer
+from mercury_engine_data_structures.formats import Bmsmsd
+
+from open_samus_returns_rando.constants import ALL_AREAS
+from open_samus_returns_rando.patcher_editor import PatcherEditor
+
+
+def _patch_tiles(tiles: ListContainer, editor: PatcherEditor):
```
ThanatosGit, line 8: Remove the first parameter. It's not used at all

ThanatosGit (approved):

PR #171, opened 2023-10-27
Disable `SetLowModelsVisibility` for samus during boss fights

ThanatosGit:
I don't really know what this is doing.
Does Samus has a different model / mesh / hitbox / whatever during the boss fights?

Following the logic during, entering a boss fights makes `C01_Body_Ch` not visible but `C01_Combat_Ch`. If you die or beat the boss, it's flipped again.
And with only varia it looks like there is no `C01_Combat_Ch` and that's why samus is invisible.

Beat ridley with varia and this patch. If it breaks something with gravity suit on, we maybe need to make it check for the suit etc.

### DIFF
--- a/open_samus_returns_rando/files/levels/s070_area7.lua
+++ b/open_samus_returns_rando/files/levels/s070_area7.lua
@@ -835,10 +835,10 @@ function s070_area7.OnPlayerDead(_ARG_0_)
   s070_area7.SetLowModelsVisibility(false)
 end
 function s070_area7.SetLowModelsVisibility(_ARG_0_)
-  if Game.GetEntity("Samus") ~= nil then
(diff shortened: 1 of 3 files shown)


Miepee:
DCR mentioned that Diggernaut, queen and Ridley do that
all 3 of the bosses have some zoomout shenanigans during their fight
maybe the game, due to zoom out and the very low 3ds screen, then uses a low-poly model of samus instead?

dyceron:
That's makes sense why non gravity is invisible because they didn't bother making a low poly model for those suits

dyceron:
That's makes sense why non gravity is invisible because they didn't bother making a low poly model for those suits

dyceron:
That's makes sense why non gravity is invisible because they didn't bother making a low poly model for those suits

dyceron (approved):

PR #172, opened 2023-10-30
Remove Grapple Block below A7 Chozo Seal

dyceron:
Removes a softlock without Grapple


ThanatosGit (approved):
It's this grapple block, right?  Very weird softlock, so good to remove it.
image

dyceron:
Yes that's the one

PR #177, opened 2023-11-09
Fix map icons function

dyceron:
Now accounts for all door changes and is more simplified

### DIFF
--- a/open_samus_returns_rando/specific_patches/map_icons.py
+++ b/open_samus_returns_rando/specific_patches/map_icons.py
@@ -31,19 +31,19 @@ def patch_tiles(editor: PatcherEditor):
 
                 icons = current_tile["icons"]
                 if len(icons) != 0:
-                    icons_length = icons[0] if len(icons) == 1 else icons[1]
+                    icon_idx = icons[0] if len(icons) == 1 else icons[1]
                     # Items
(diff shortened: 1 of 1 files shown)


open_samus_returns_rando/specific_patches/map_icons.py
```diff
@@ -29,24 +29,21 @@ def patch_tiles(editor: PatcherEditor):
...
-                    else:
-                        icons[0]["icon"] = "itemenabled"
-
-                for door in doors:
-                    if len(icons) == 1 and icons[0]["actor_name"].startswith(door):
-                        icons[0]["clear_condition"] = ""
-                        if "left" in icons[0]["icon"]:
-                            icons[0]["icon"] = "doorpowerleft"
+                if len(icons) != 0:
+                    icons_length = icons[0] if len(icons) == 1 else icons[1]
```
ThanatosGit, line 34: Can we rename that variable? It is very confusing for me to name something `...length` and it is not a number but a dict instead. (It's cleary not a `length` at all?!)

ThanatosGit (changes requested):

open_samus_returns_rando/specific_patches/map_icons.py
dyceron: Sure

ThanatosGit (approved):

PR #178, opened 2023-11-09
Fix Area 8 Larvas

dyceron:
- Removes the Larva intro cutscene
- Adds 10 Metroids to the Metroid counter (should this only be added on when in Area 8? Would this cause confusion in routing?)
- Prevents a door from locking if entering the room with Larva_005 to remove a softlock

### DIFF
--- a/open_samus_returns_rando/files/levels/s100_area10.lua
+++ b/open_samus_returns_rando/files/levels/s100_area10.lua
@@ -146,7 +146,7 @@ function s100_area10.OnLarva_005_Generated(_ARG_0_, _ARG_1_)
   if _ARG_1_ ~= nil and _ARG_1_.AI ~= nil then
     _ARG_1_.AI.bPlaceholder = false
     _ARG_1_.AI:AddPresetLogicPath(0, "PATH_Larva_005")
-    _ARG_1_.AI:AddDoorToLock("Door013")
+    -- _ARG_1_.AI:AddDoorToLock("Door013")
     _ARG_1_.AI:AddDoorToLock("Door014")
     _ARG_1_.AI:AddSpawnPointToUnlock("Larva_005")
     _ARG_1_.AI:AddSpawnPointToUnlock("Larva_006")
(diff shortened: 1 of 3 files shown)


ThanatosGit (approved):
> Adds 10 Metroids to the Metroid counter (should this only be added on when in Area 8? Would this cause confusion in routing?)

It depends. I never played with the queen wall removed. Are the metroids present without it or are they only present with the `MetroidDiscovered` flag set to true via this PR?

If it is confusing, the less confusing thing would probably be to just leave the cutscene in and do the whole vanilla incrementing metroid counter thing, so a player cleary knows what happened. If going area 8 backwards is or will be a thing, this of course does not apply.

dyceron:
> > Adds 10 Metroids to the Metroid counter (should this only be added on when in Area 8? Would this cause confusion in routing?)
> 
> It depends. I never played with the queen wall removed. Are the metroids present without it or are they only present with the `MetroidDiscovered` flag set to true via this PR?
> 
> If it is confusing, the less confusing thing would probably be to just leave the cutscene in and do the whole vanilla incrementing metroid counter thing, so a player cleary knows what happened. If going area 8 backwards is or will be a thing, this of course does not apply.
> 

I honestly don't know if setting that prop changes anything here. The Metroids always spawn I'm pretty sure.

I could change this so the cutscene only plays if entering via the tunnel, but that doesn't solve the metroid count issue. Which is kinda why I made the counter higher by default. I think I should change this so the counter increments regardless which side is entered.

ThanatosGit:
> I could change this so the cutscene only plays if entering via the tunnel, but that doesn't solve the metroid count issue. Which is kinda why I made the counter higher by default. I think I should change this so the counter increments regardless which side is entered.

What do you think about to keep the Count at 40, so you know how many metroids still locks items?
And keep the removal of the cutscene and removal of the softlock.

dyceron:
Could do that

dyceron:
> What do you think about to keep the Count at 40, so you know how many metroids still locks items? And keep the removal of the cutscene and removal of the softlock.

I haven't quite figured out how to not make the Metroids decrement the counter on death, otherwise I would change the total back to 40 again

ThanatosGit:
Oh, good point. Then leave it as it is.

dyceron:
> Oh, good point. Then leave it as it is.

PR should be good to go then. Later on we can adjust this by fixing the counter or making it increase when entering the area/set of cameras. I've thought about having the counter update per area to show how many metroids are left, but that's for another time.

PR #183, opened 2023-11-23
Prevent collecting DIggernaut's item during cutscene

dyceron:


ThanatosGit (approved):
Ah, on cutscene start. That's definitely a better place than what I suggested.

PR #185, opened 2023-11-23
Create bmsad per item and reuse them instead of creating a bmsad for each location

ThanatosGit:
Probably much better in memory consumption to use one bmsad per item instead of per location. That's reducing the required custom bmsads by something like over 100 files.

(Should be rebased later. Accidentally, have the chozo seal dna stuff in here)

### DIFF
--- a/open_samus_returns_rando/pickup.py
+++ b/open_samus_returns_rando/pickup.py
@@ -6,7 +6,7 @@ from enum import Enum
 from construct import Container, ListContainer
 from mercury_engine_data_structures.formats import Bmsad
 
-from open_samus_returns_rando.constants import get_package_name, lua_pkgs
+from open_samus_returns_rando.constants import get_package_name
 from open_samus_returns_rando.files import templates_path
 from open_samus_returns_rando.logger import LOG
 from open_samus_returns_rando.lua_editor import LuaEditor
@@ -85,6 +85,8 @@ class BasePickup:
         raise NotImplementedError
 
 class ActorPickup(BasePickup):
+    _bmsad_dict = {}
+
     def patch_item_pickup(self, bmsad: dict) -> dict:
         pickable: dict = bmsad["components"]["PICKABLE"]
         script: dict = bmsad["components"]["SCRIPT"]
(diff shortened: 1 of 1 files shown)


dyceron (approved):

open_samus_returns_rando/pickup.py
```diff
@@ -262,10 +286,7 @@ def count_dna(lua_scripts: LuaEditor, pickup_object: BasePickup):
         lua_scripts.add_dna(scenario)
 
 def patch_pickups(editor: PatcherEditor, lua_scripts: LuaEditor, pickups_config: list[dict], configuration: dict):
-    all_pkgs = editor.get_all_level_pkgs()
-    editor.add_new_asset("actors/items/randomizer_powerup/scripts/randomizer_powerup.lc", b'', lua_pkgs(all_pkgs))
-    # FIXME: Why it doesn't work when we directly add it to the pkgs?
-    # Addition: Tested => It seems like this file is read fro RomFS but why not from pkg?
+    editor.add_new_asset("actors/items/randomizer_powerup/scripts/randomizer_powerup.lc", b'', [])
```
ThanatosGit, line 289: This still looks a little bit weird but some lua files were not read from pkg files, so I decided to not add our own files to the pkgs at all.

open_samus_returns_rando/pickup.py
```diff
@@ -146,7 +169,11 @@ def patch_model(self, model_names: list[str], bmsad: dict) -> None:
                 fx_create_and_link["Param1"]["value"] = "leak"
                 fx_create_and_link["Param2"]["value"] = "actors/items/adn/fx/adnleak.bcptl"
                 fx_create_and_link["Param13"]["value"] = True
+            else:
+                bmsad["components"].pop("FX")
+                bmsad["sound_fx"] = ListContainer([])
         else:
+            bmsad["components"].pop("FX")
```
dyceron, line 176: I wonder if we could add this here too `bmsad["sound_fx"] = ListContainer([])`

ThanatosGit: In vanilla that's the case for majors iirc. Minors (at least missile tanks) have the stuff in there which is in our template_bmsad.json

PR #186, opened 2023-11-23
Fix unmapped reads / writes

ThanatosGit:
We need to check for `SpawnGroup020` and `SpawnGroup021` because I'm sure that this was throwing also errors when very close to the door,

Should also be rebased later.

### DIFF
--- a/open_samus_returns_rando/files/levels/s040_area4.lua
+++ b/open_samus_returns_rando/files/levels/s040_area4.lua
@@ -206,9 +206,9 @@ function s040_area4.OnEnter_Gamma_001_Dead()
     Game.SetSubAreaCurrentSetup("collision_camera_005", "PostGamma_001", true)
     Game.SetSubAreaCurrentSetup("collision_camera_012", "PostGamma_001", true)
     Game.SetSubAreaCurrentSetup("collision_camera_013", "PostGamma_001", true)
-    if Game.GetEntity("SpawnGroup019") ~= nil then
-      Game.GetEntity("SpawnGroup019").SPAWNGROUP:EnableSpawnGroup()
-    end
+    -- if Game.GetEntity("SpawnGroup019") ~= nil then
+      -- Game.GetEntity("SpawnGroup019").SPAWNGROUP:EnableSpawnGroup()
+    -- end
     if Game.GetEntity("SpawnGroup020") ~= nil then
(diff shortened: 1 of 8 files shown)


open_samus_returns_rando/specific_patches/door_patches.py
```diff
@@ -98,8 +97,7 @@ def _patch_beam_bmsads(editor: PatcherEditor):
         editor.replace_asset(creature_bmsad_file, cr_bmsad)
 
     editor.add_new_asset(
-        "actors/props/doors/scripts/doors.lc", files_path().joinpath("doors.lua").read_bytes(),
-        lua_pkgs(editor.get_all_level_pkgs())
+        "actors/props/doors/scripts/doors.lc", files_path().joinpath("doors.lua").read_bytes(), []
```
ThanatosGit, line 100: Same thing here as in the other PR: Moved lua files to the RomFS because they are read from there without issues and some didn't work in the pkg.

open_samus_returns_rando/files/levels/s040_area4.lua
```diff
@@ -649,5 +649,12 @@ function s040_area4.OnExit_ChangeCamera_023()
   Game.SetCollisionCameraLocked("collision_camera_023_B", false)
 end
 function s040_area4.OnSubAreaChange(_ARG_0_, _ARG_1_, _ARG_2_, _ARG_3_, _ARG_4_)
+  -- delay activation of this SG because it throws memory errors if you stay at the
+  -- area where the gamma metroid dies
+  if _ARG_0_ ~= "collision_camera_013" and Scenario.ReadFromBlackboard("Arena_Gamma_001_AllDead", false) then
+    if Game.GetEntity("SpawnGroup019") ~= nil then
+      Game.GetEntity("SpawnGroup019").SPAWNGROUP:EnableSpawnGroup()
+    end
+  end
```
ThanatosGit, line 658: Here is the part we should verify on citra and console that it works without including sg 20 and sg 21

dyceron, line 658: We can revisit this later if the crashes are still around

ThanatosGit: Fair. At least, I'm pretty sure if the metroid rooms is crashing that we know why and can easily fix it.

PR #188, opened 2023-11-23
Match vanilla files and fix surface connection

dyceron:


open_samus_returns_rando/files/levels/s000_surface.lua
```diff
@@ -459,10 +459,9 @@ function s000_surface.OnSubAreaChange(_ARG_0_, _ARG_1_, _ARG_2_, _ARG_3_, _ARG_4
   elseif (_ARG_0_ == "collision_camera_024" and not Scenario.ReadFromBlackboard("alpha_killed", false)) then
     s000_surface.LaunchFirstTimeAlphaPresentation()
   end
-  if _ARG_0_ == "collision_camera_000" and _ARG_2_ == "collision_camera_017" then
+  if _ARG_2_ == "collision_camera_017" then
```
dyceron, line 462: actually this should probably use arg0 instead

ThanatosGit: What's the current issue with the code?
IIRC _ARG_0 is where you came from and _ARG_2_ where you are going?

dyceron: The idea is to enable the "warp" no matter what cc you approach from. Really only relevant for min logic/oob nonsense. This ensures that surface west always gets loaded.

ThanatosGit: Ah, okay. Then I guess it's fine.

ThanatosGit (approved):

PR #189, opened 2023-11-23
Empty the static dict when start patching.

ThanatosGit:
The dict keeps being filled if you try to make more than one export in a row via randovania because it is static.
So, I clear it before loop through the pickups.


dyceron (approved):
Fix conflicts

PR #190, opened 2023-11-23
Fix map tiles

dyceron:
Support edge case where len is 2 and first object is a tank


ThanatosGit (approved):
No idea where this is case is but I assume it's right :D

dyceron:
> No idea where this is case is but I assume it's right :D

Items in the same tile as doors.

PR #197, opened 2023-11-26
Update README.md

dyceron:


ThanatosGit:
That reminds me of: Do we want to split the installation section?
The `pip install open-samus-returns-rando` will be valid if you want to use the patcher outside of randovania (and when we release a version), while the `pre-commit` part is actually about developing.
For developing it should be just `pip install .`  + the pre-commit part?

dyceron:
> That reminds me of: Do we want to split the installation section? The `pip install open-samus-returns-rando` will be valid if you want to use the patcher outside of randovania (and when we release a version), while the `pre-commit` part is actually about developing. For developing it should be just `pip install .` + the pre-commit part?

Yeah we can

ThanatosGit (approved):

PR #205, opened 2023-12-07
Pickup files cleanup and improvements

dyceron:
Aims to make each pickup bmsad be as accurate as possible, as well as add anything else missing for the pickups to function properly


dyceron:
I can't think of anything else atm so I'm marking it as ready for review. My only thing is if we even need to ensure or bother with sound fx at all, or if all of the items are using the correct sound fx component in the custom bmsads (assuming they are used at all)

ThanatosGit (approved):

PR #223, opened 2023-12-17
Disable PBs during Diggernaut fight

dyceron:
Sets PB MAX to 0 while the pickup is disabled and restores the value when the pickup is re-enabled


ThanatosGit (changes requested):
The solution is not fool proof for multiworld.
1. pbs will get activated if you receive a pb tank during this long fight
2. it will also reset the amount of tanks to pre-fight value and received one's during the fight are lost.

At least the second should be addressed.

dyceron:
> The solution is not fool proof for multiworld.
> 
>     1. pbs will get activated if you receive a pb tank during this long fight
> 
>     2. it will also reset the amount of tanks to pre-fight value and received one's during the fight are lost.
> 
> 
> At least the second should be addressed.

Yeah I was thinking about multiworld. Theres a field for diggernaut which has pbs set to true, but making it false does nothing. its the actual wall I think. But I even disabled the wall and it still ended, so this needs more thought.

ThanatosGit:
> Theres a field for diggernaut which has pbs set to true, but making it false does nothing. its the actual wall I think.

Diggernaut should be the `manicminerbot.bmsad`. If changing it doesn't work, it's probably buried somewhere deeper. Which very likely means, we can't change it back after defeating diggernaut anyways.

Just address the 2. point and it is fine.

dyceron:
I changed this so now the value of launcher and max are stored before the fight and are restored afterwards. The launcher is stored and disabled so getting PBs during the fight doesn't allow for using them. I also keep track of how many PBs were recieved during the fight (multiworld only) and add that value to the max after.

src/open_samus_returns_rando/files/levels/s070_area7.lua
```diff
@@ -889,6 +893,9 @@ function s070_area7.OnStartManicMinerBotDeathCutscene()
   Game.AddEntityToUpdateInCutscene("LE_ManicMinerWall")
   Game.AddEntityToUpdateInCutscene("LE_PowerUp_Powerbomb")
   Scenario.WriteToBlackboard("manicMinerDead", "b", true)
+  local PostFightPBMax = Blackboard.GetProp("PLAYER_INVENTORY", "ITEM_WEAPON_POWER_BOMB_MAX")
+  Blackboard.SetProp("PLAYER_INVENTORY", "ITEM_WEAPON_POWER_BOMB", "f", PBLauncher)
+  Blackboard.SetProp("PLAYER_INVENTORY", "ITEM_WEAPON_POWER_BOMB_MAX", "f", PostFightPBMax + PreFightPBMax)
```
ThanatosGit, line 898: The launcher does not exist in the vanilla game like that. As soon as we have `ITEM_WEAPON_POWER_BOMB_MAX > 0` you can use PBs.

What item we grant is modified depending if `ITEM_WEAPON_POWER_BOMB` is in the inventory or not.
The item sent or picked up is always `ITEM_POWER_BOMB_TANKS`. (no idea why the plural made it into the code).
So, it is kept as `ITEM_POWER_BOMB_TANKS` if `ITEM_WEAPON_POWER_BOMB = 0` else it is changed to `ITEM_WEAPON_POWER_BOMB_MAX `.
See:
https://github.com/randovania/open-samus-returns-rando/blob/195969116cce52f8ac7de594d8832997ee7d9716/src/open_samus_returns_rando/files/templates/randomizer_powerup.lua#L195-L200

That makes the whole logic cursed especially if you take into account you could receive the main item during the fight (or a second one).

AFAIK you need to get `ITEM_WEAPON_POWER_BOMB` after the fight and sum it up with `PBLauncher`. If that is ` >=1` we either had the launcher before or received it during the fight. 
`PostFightPBMax` needs to be summed up with with `ITEM_POWER_BOMB_TANKS` (a.k.a. "locked" PB tanks) and line 898 needs to be conditional: 
1. Use `ITEM_WEAPON_POWER_BOMB_MAX` if `"sum of launchers" >= 1` and reset `ITEM_POWER_BOMB_TANKS` to 0
2. else keep them locked and use  `ITEM_POWER_BOMB_TANKS` .

ThanatosGit (changes requested):
Damn it. After reading your text and the code, I just realised that this is wrong.

src/open_samus_returns_rando/files/levels/s070_area7.lua
dyceron: Wouldn't getting a Launcher mid

ThanatosGit: Yeah, it would break in the meaning of that the PBs gets activated.
But it would also break much more: The `ITEM_WEAPON_POWER_BOMB` would be set to pre-fight value, which was 0. Due to the linked logic from above, you won't get any new tanks because if you pick them up, they will be locked as you don't have the main item.

Example:
```
---
Pre-fight
ITEM_WEAPON_POWER_BOMB = 0,
ITEM_WEAPON_POWER_BOMB_MAX = 0,
ITEM_POWER_BOMB_TANKS = 5,
PBLauncher = 0
PreFightPBMax = 0
---

---
During fight (receive main item => unlocks the tanks):
ITEM_WEAPON_POWER_BOMB = 1,
ITEM_WEAPON_POWER_BOMB_MAX = 5,
ITEM_POWER_BOMB_TANKS = 0,
PBLauncher = 0
PreFightPBMax = 0
---

---
After fight:
ITEM_WEAPON_POWER_BOMB = 0, (was reset to PBLauncher value)
ITEM_WEAPON_POWER_BOMB_MAX = 5,
ITEM_POWER_BOMB_TANKS = 0,
PBLauncher = 0
PreFightPBMax = 0

=> 5 PBs are usable, no more are added because main item was reset to 0
---
```

dyceron: Oh I see the issue. So I should change both or just the launcher then?

ThanatosGit: Don't know what you mean by "both" here. But basically it should be this(?):

> AFAIK you need to get `ITEM_WEAPON_POWER_BOMB` after the fight and sum it up with `PBLauncher`. If that is ` >=1` we either had the launcher before or received it during the fight. `PostFightPBMax` needs to be summed up with with `ITEM_POWER_BOMB_TANKS` (a.k.a. "locked" PB tanks) and line 898 needs to be conditional:
> 
> 1. Use `ITEM_WEAPON_POWER_BOMB_MAX` if `"sum of launchers" >= 1` and reset `ITEM_POWER_BOMB_TANKS` to 0
> 2. else keep them locked and use  `ITEM_POWER_BOMB_TANKS` .

dyceron: I hope this is what you meant

dyceron:
I believe these changes should cover the important bases. 

Before the fight:
- Store the inventory value of the launcher
 - If we don't have the launcher, ignore conditional and continue keeping track of PBs as tanks
 - If we do have the launcher, set the value of `_MAX` to a variable and set launcher and `_MAX` to 0 (disable pbs). Then set the tanks value to be the stored `_MAX` value

After the fight, check for the value of PBs and add it to the stored value from before:
- If we didn't have the launcher before the fight and didn't receive it mid-fight, ignore conditional and continue keeping track of PBs as tanks
- If we had the launcher before the fight or received it mid-fight:
 - Set PBs to 1 (has main)
 - Set `_MAX` to the value of tanks
 - Reset tanks to 0

src/open_samus_returns_rando/files/levels/s070_area7.lua
```diff
@@ -889,6 +896,11 @@ function s070_area7.OnStartManicMinerBotDeathCutscene()
   Game.AddEntityToUpdateInCutscene("LE_ManicMinerWall")
   Game.AddEntityToUpdateInCutscene("LE_PowerUp_Powerbomb")
   Scenario.WriteToBlackboard("manicMinerDead", "b", true)
+  if PBLauncher + Game.GetItemAmount(Game.GetPlayerName(), "ITEM_WEAPON_POWER_BOMB") > 0 then
+    Game.SetItemAmount(Game.GetPlayerName(), "ITEM_WEAPON_POWER_BOMB", 1)
+    Game.SetItemAmount(Game.GetPlayerName(), "ITEM_WEAPON_POWER_BOMB_MAX", "ITEM_POWER_BOMB_TANKS")
```
ThanatosGit, line 901: I think there are two issues here:
1. I only used `Game.SetItemAmount` with a number as third parameter. `"ITEM_POWER_BOMB_TANKS"` is just a string.

2. You are now using `"ITEM_POWER_BOMB_TANKS"` to store `PreFightPBMax`. Which I think is a very good idea but then you need to change the if clause in line 899.
If you receive the Launcher mid fight, all of the locked `ITEM_POWER_BOMB_TANKS` will be unlocked by our Powerbomb pickup logic. In that case you don't need to anything after the fight.
The only case you need to do something is if the main item was removed pre fight and needs to be unlocked post fight.
See:

https://github.com/randovania/open-samus-returns-rando/blob/ec559eabd04df7071c602e2bba0cdf471f5125ad/src/open_samus_returns_rando/files/templates/randomizer_powerup.lua#L195-L210

src/open_samus_returns_rando/files/levels/s070_area7.lua
```diff
@@ -889,6 +896,11 @@ function s070_area7.OnStartManicMinerBotDeathCutscene()
   Game.AddEntityToUpdateInCutscene("LE_ManicMinerWall")
   Game.AddEntityToUpdateInCutscene("LE_PowerUp_Powerbomb")
   Scenario.WriteToBlackboard("manicMinerDead", "b", true)
+  if PBLauncher + Game.GetItemAmount(Game.GetPlayerName(), "ITEM_WEAPON_POWER_BOMB") > 0 then
```
ThanatosGit, line 899:
```suggestion
  if PBLauncher ~= 0 and Game.GetItemAmount(Game.GetPlayerName(), "ITEM_WEAPON_POWER_BOMB") == 0 then
```

src/open_samus_returns_rando/files/levels/s070_area7.lua
ThanatosGit, line 901:
```suggestion
    local locked_pbs = Game.GetItemAmount(Game.GetPlayerName(), "ITEM_POWER_BOMB_TANKS")
    Game.SetItemAmount(Game.GetPlayerName(), "ITEM_WEAPON_POWER_BOMB_MAX", locked_pbs )
```

ThanatosGit (changes requested):

src/open_samus_returns_rando/files/levels/s070_area7.lua
dyceron: 1. Good catch
2. That makes sense since the other case is already handled in our code

ThanatosGit (approved):

PR #232, opened 2023-12-23
Infinite load stuff

ThanatosGit:
Before merge / making a release out of it, the annotated stuff needs to be addressed.


src/open_samus_returns_rando/files/custom_ship.lua
```diff
@@ -6,7 +6,7 @@ end
 function SamusShip.Dummy()
 end
 function SamusShip.WarpToStart()
-  Scenario.LoadNewScenario(Init.sStartingScenario, Init.sStartingActor)
```
ThanatosGit, line 9: revert that. this is only for reproducible example

tests/test_files/rdv_seed.json
ThanatosGit, line 1: This file can be removed and is also only for the example

dyceron (approved):

PR #238, opened 2024-01-01
Improve disabling liquid actors if collecting Gravity

dyceron:
- The suit model now properly updates when suits are not progressive (Power -> Gravity now uses Gravity model instead of Varia)
- The check for enabling/disabling triggers has been overhauled
  - Adds in some triggers that were unaccounted for
  - Combines both functions into a more streamlined one
  - Checks if Samus is inside a trigger and only disables that specific one instead of disabling all of them in the scenario
  - Accounts for all liquids (acid, lava, water) regardless if they normally have an item or not
    - This only applies to Multiworld, as it could be possible to receive Gravity while in one of these triggers which would force the player to reload their save
- Removes unused/uncessary function calls elsewhere

### DIFF
--- a/src/open_samus_returns_rando/files/templates/randomizer_powerup.lua
+++ b/src/open_samus_returns_rando/files/templates/randomizer_powerup.lua
@@ -73,11 +73,9 @@ if RandomizerPowerup == nil then
             -- both loops cover repeated/similar actor names that are shared across scenarios
             if scenario ~= "s000_surface" and scenario ~= "s010_area1" and scenario ~= "s110_surfaceb" then
                 local waterPrefixes = {"TG_Water_", "TG_SP_Water_"}
-                local waterNumbers = {"001", "002", "003", "004", "005", "006", "007", "008", "009", "010", "011", "012",
-                    "013", "014", "015", "016", "017", "018", "019"}
                 for i = 1, 2 do
                     for j = 1, 19 do
-                        local water = Game.GetEntity(waterPrefixes[i] .. waterNumbers[j])
+                        local water = Game.GetEntity(waterPrefixes[i] .. string.format("%03d", j))
                         if water ~= nil and water.TRIGGER:IsPlayerInside() == true then
                             water.TRIGGER:DisableTrigger()
                             water.TRIGGER:EnableTrigger()
@@ -89,8 +87,7 @@ if RandomizerPowerup == nil then
                 local lavaScenarios = {"s020_area2", "s033_area3b", "s036_area3c", "s040_area4", "s050_area5"}
                 if scenario == lavaScenarios[i] then
                     for j = 1, 5 do
-                        local lavaNumbers = {"001", "002", "003", "004", "005"}
-                        lava = Game.GetEntity("TG_Lava_" .. lavaNumbers[j])
+                        lava = Game.GetEntity("TG_Lava_" .. string.format("%03d", j))
                         if lava ~= nil and lava.TRIGGER:IsPlayerInside() == true then
                             lava.TRIGGER:DisableTrigger()
                             lava.TRIGGER:EnableTrigger()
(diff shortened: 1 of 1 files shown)


ThanatosGit (commented):
Looks fine to me.
My only question is: Do we want to keep the amount of lua code as low as possible due to the ils stuff?

dyceron:
> Looks fine to me. My only question is: Do we want to keep the amount of lua code as low as possible due to the ils stuff?

With split files, idk if this would be a concern or not. I mainly was verbose since I felt it was unnecessary/wasteful to have a full array of actors and check them against each scenario when some of these don't exist in all of them. But, if its actually not a big deal to have a full list to check from then I could simplify this into combined loops.

src/open_samus_returns_rando/files/templates/randomizer_powerup.lua
```diff
@@ -64,27 +64,71 @@ if RandomizerPowerup == nil then
...
-                Game.GetEntity("TG_SP_Water_002").TRIGGER:DisableTrigger()
+            local scenario = Scenario.CurrentScenarioID
+            -- both loops cover repeated/similar actor names that are shared across scenarios
+            if scenario ~= "s000_surface" and scenario ~= "s010_area1" and scenario ~= "s110_surfaceb" then
+                local waterPrefixes = {"TG_Water_", "TG_SP_Water_"}
+                local waterNumbers = {"001", "002", "003", "004", "005", "006", "007", "008", "009", "010", "011", "012",
+                    "013", "014", "015", "016", "017", "018", "019"}
+                for i = 1, 2 do
+                    for j = 1, 19 do
+                        local water = Game.GetEntity(waterPrefixes[i] .. waterNumbers[j])
```
ThanatosGit, line 80: Remove the whole `waterNumbers` thing. What you were looking for is a format string.
Replace `waterNumbers[j]` by `string.format("%03d", j)`.

src/open_samus_returns_rando/files/templates/randomizer_powerup.lua
```diff
@@ -64,27 +64,71 @@ if RandomizerPowerup == nil then
...
+                        end
+                    end
+                end
+            end
+            for i = 1, 5 do
+                local lavaScenarios = {"s020_area2", "s033_area3b", "s036_area3c", "s040_area4", "s050_area5"}
+                if scenario == lavaScenarios[i] then
+                    for j = 1, 5 do
+                        local lavaNumbers = {"001", "002", "003", "004", "005"}
+                        lava = Game.GetEntity("TG_Lava_" .. lavaNumbers[j])
```
ThanatosGit, line 93: Same as for `waterNumbers`

src/open_samus_returns_rando/files/templates/randomizer_powerup.lua
```diff
@@ -64,27 +64,71 @@ if RandomizerPowerup == nil then
...
+                    "TG_HazardousPuddle_004", "TG_HazardousPuddle_005"}
+                arrayLength = 5
+            elseif scenario == "s065_area6b" then
+                liquids = {"TG_LS_Water_001"}
+                arrayLength = 1
+            elseif scenario == "s070_area7" then
+                liquids = {"TG_Damage_Hazardous_001", "TG_Damage_Hazardous_002"}
+                arrayLength = 1
+            end
+            for i = 1, arrayLength do
```
ThanatosGit, line 125: Remove the `arrayLength` thing. `#liquids` gives you the length. It's like `len(liquids)` in python.

ThanatosGit (changes requested):

src/open_samus_returns_rando/files/templates/randomizer_powerup.lua
dyceron: So remove the variable altogether?

ThanatosGit: Yeah. I mean, nothing of what you did (or in the other comments) is wrong, it just falls kinda into the "reinventing the wheel" because it can be done with built-in functions from lua itself.

ThanatosGit (approved):

PR #239, opened 2024-01-02
Support Shuffling Missiles

dyceron:
- Allows using Supers without having Missiles
- Adds Missile Launcher pickup
- Reorganizes some files

image


src/open_samus_returns_rando/pickups/pickup.py
```diff
@@ -215,6 +216,9 @@ def patch_model(self, model_names: list[str], bmsad: dict) -> None:
                 MODELUPDATER["functions"][0]["params"].pop("Param2")
                 bmsad["components"].pop("FX")
                 bmsad["sound_fx"] = ListContainer([])
+            elif model_name == "powerup_missilelauncher":
+                bmsad["header"]["unk_2d"] = 1.75
```
ThanatosGit, line 220: what is this unk? You know the meaning of it? 👀

src/open_samus_returns_rando/files/templates/randomizer_powerup.lua
ThanatosGit, line 1: Git thinks you changed all these line which is horrible to review. But it looks like this part is just a copy paste from what we already had for pb and sm stuff which totally makes sense.

ThanatosGit (approved):

src/open_samus_returns_rando/files/templates/randomizer_powerup.lua
dyceron: I copy pasted but also rearranged the order so it's missiles, Supers, then pbs, though that's kinda pointless since it'll be split anyway

PR #240, opened 2024-01-03
Use less obstructive fx for Reserve Tanks

dyceron:
`spinattack.bcptl` is just cooler :)

https://github.com/randovania/open-samus-returns-rando/assets/38679103/e36ea666-9366-42bf-806f-aac40555076e


ThanatosGit:
The change itself looks fine but do we need to assure that this bcptl file is present in the scenario or is it already in every pkg? (asking because there is no change in the model_data file)

dyceron:
> The change itself looks fine but do we need to ensure that this bcptl file is present in the scenario or is it already in every pkg? (asking because there is no change in the model_data file)

I didn't actually test other scenarios but my guess is it always exists due to being tied to samus directly.

ThanatosGit (approved):

PR #244, opened 2024-01-05
Do full test on pull request and require it before pypi push

ThanatosGit:


ThanatosGit:
Should be good now. The commit messages describes it already good enough.

dyceron (approved):

PR #246, opened 2024-01-07
Add option to require Beam Burst for Blobthrowers/Blockingplants

dyceron:


src/open_samus_returns_rando/files/schema.json
```diff
@@ -180,6 +180,11 @@
                     "description": "Remove the Power Bomb weakness from Beam Doors to make Power Bombs less overpowered.",
                     "default": true
                 },
+                "plant_buff": {
+                    "type": "boolean",
+                    "description": "Remove the Power Bomb weakness from Blobthowers/Blockingplants to make Power Bombs less overpowered.",
+                    "default": true
```
ThanatosGit, line 186: My first thought was that I don't really want this to default to true. On a second thought, it could be a good reason to not give you early PBs, so that you can enter Area 3 with something different.
Now, I don't know what I like 😐

dyceron: I think default is fine. Could always just swap in rdv itself

ThanatosGit: But it makes current patcher files unbeatable. So, this is version 2.0.0 😛?

ThanatosGit (approved):

PR #247, opened 2024-01-08
Nerf beams against ridley

ThanatosGit:
Don't know when or if I will test it. But we have beta testers :)
(raising it for wave beam and plasma beam was working)

### DIFF
--- a/src/open_samus_returns_rando/specific_patches/static_fixes.py
+++ b/src/open_samus_returns_rando/specific_patches/static_fixes.py
@@ -79,6 +79,18 @@ def shoot_supers_without_missiles(editor: PatcherEditor):
     gun[20]["params"]["Param5"]["value"] = ""
 
 
+def nerf_ridley_fight(editor: PatcherEditor):
+    # Power Beam: 3 (25 * 0.12)
+    # Ice: 10
+    # Wave: 14 (50 * 0.28)
+    # Spazer: 25,2 (210 * 0.12)
+    # Plasma: 36 (default of 300 * 0.12 in tunable of ridley)
+    ridley_bmsad = editor.get_file("actors/characters/ridley/charclasses/ridley.bmsad", Bmsad)
+    ridley_bmsad.components["LIFE"].raw["fields"]["fPowerBeamFactor"]["value"] = 0.12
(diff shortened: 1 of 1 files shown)


dyceron (commented):
Do we not have worry about charge? Also what about ice? It ends up being stronger than wave and power now

ThanatosGit:
> Do we not have worry about charge?
See:
> Don't know when or if I will test it. But we have beta testers :)

All bullets should be:
Power Beam: 3 (25 * 0,12)
Ice: 10
Wave: 6 (50 *0,12)
Spazer: 25,2 (210 * 0,12)
Plasma: 36 (default of 12 * 3)

So, we make Wave to be 14 or 15?

dyceron:
> All bullets should be:
> Power Beam: 3 (25 * 0,12)
> Ice: 10
> Wave: 6 (50 *0,12)
> Spazer: 25,2 (210 * 0,12)
> Plasma: 36 (default of 12 * 3)
> 
> So, we make Wave to be 14 or 15?
> 

We could have Wave be 0.17 which is 8.5 damage or 1/3 of spazer (literally makes sense). Following that, Power could be 4.25 (0.17) or half of Wave as its normally half the strength of Wave, then have Ice use 0.33 or 3.33 damage

src/open_samus_returns_rando/specific_patches/static_fixes.py
```diff
@@ -79,10 +79,18 @@ def shoot_supers_without_missiles(editor: PatcherEditor):
     gun[20]["params"]["Param5"]["value"] = ""
 
 
+def nerf_ridley_fight(editor: PatcherEditor):
+    ridley_bmsad = editor.get_file("actors/characters/ridley/charclasses/ridley.bmsad", Bmsad)
+    ridley_bmsad.components["LIFE"].raw["fields"]["fSpazerBeamFactor"]["value"] = 0.12
+    ridley_bmsad.components["LIFE"].raw["fields"]["fWaveBeamFactor"]["value"] = 0.12
```
dyceron, line 85: I think we decided on 14? Change it 0.28.

dyceron (changes requested):

dyceron (approved):

PR #251, opened 2024-01-10
Refactoring for seperated powerup files and byte code for lua files via new MEDS

ThanatosGit:


dyceron (approved):
Some of this stuff is confusing to me (as per usual 😛) but it looks good. Everything seems to function fine in game as it did before.

PR #259, opened 2024-01-24
Support Cosmetic Patches

dyceron:
First batch of cosmetics. This is pretty good for now and not too overbearing. This does not implement the Energy Counter text as there are issues with keeping it active that I haven't worked out. This supports changing:
- Free-Aim color (locked and unlocked)
- Grapple Aim color (locked and unlocked)
- Energy Tank color
- Aeion Bar color
- Ammo HUD color

Locked Aim
image

Unlocked Aim
image

Locked Grapple
image

Unlocked Grapple
image

Example HUD
image


src/open_samus_returns_rando/lua_editor.py
```diff
@@ -284,6 +285,19 @@ def save_modifications(self, editor: PatcherEditor, configuration: dict):
...
+            Lua(Container(lua_text=files_path().joinpath("templates", "cosmetics.lua").read_text()),
+                editor.target_game),
+            []
+        )
+
+        cosmetics_script = cosmetic_patches._lua_cosmetics(configuration["cosmetic_patches"])
+        editor.replace_asset(
+            "system/scripts/cosmetics.lc",
+            Lua(Container(lua_text=cosmetics_script), editor.target_game)
+        )
```
ThanatosGit, line 299: Simplify the whole block here.
It's adding a new asset with the `TEMPLATE` variables in the lua, then gets the whole script with the replaced template and then replaces the asset with the content from the step before.

You can directly add the asset with the correct content.

src/open_samus_returns_rando/specific_patches/cosmetic_patches.py
```diff
@@ -0,0 +1,36 @@
+from mercury_engine_data_structures.formats.bmtun import Bmtun
+from open_samus_returns_rando.misc_patches import lua_util
+from open_samus_returns_rando.patcher_editor import PatcherEditor
+
+
+def patch_cosmetics(editor: PatcherEditor, configuration: dict):
+    tunables = editor.get_file("system/tunables/tunables.bmtun", Bmtun)
+    _tunable_cosmetics(tunables, configuration)
+    _lua_cosmetics(configuration)
```
ThanatosGit, line 9: This line is doing nothing. It gives you the whole lua file as a string but you never assign it to a variable and just throw it away.
It's not needed because you do that in the part where my other comment is.

ThanatosGit (changes requested):

src/open_samus_returns_rando/specific_patches/cosmetic_patches.py
dyceron: Is it not calling to the replacement below?

ThanatosGit: It's calling it but the only thing the replacement really does is returning a string with what you need to do something. As the return is not even assigned, I can easily tell this is doing (logically) nothing.

PR #261, opened 2024-01-27
Add custom chozo seals

dyceron:
Adds new Chozo Seals to areas scenarios which do not have them. These new seals will be strictly used for DNA hints. Seals have been added to the following scenarios:

area3b - Lowest Gamma+ Arena

image

area5 - Gamma Arena Lower CC

image

area6b - Top of Tower

image
 
area10 - Middle Hallway before the big room

image

area10 - After the Queen (deactivated until the Queen is defeated)

image

surfaceb - Before the Teleporter

image

Unsure how many of these I should add, or if I should add more. I think adding one more earlier on that doesn't require many items could be a good thing. Unsure about adding one in a6/a7 though.

### DIFF
--- a/src/open_samus_returns_rando/specific_patches/chozo_seal_patches.py
+++ b/src/open_samus_returns_rando/specific_patches/chozo_seal_patches.py
@@ -95,14 +95,14 @@ def add_chozo_seals(editor: PatcherEditor, new_seal: NewChozoSeal):
     scenario_name = new_seal.scenario
     scenario_file = editor.get_scenario(scenario_name)
 
-    editor.copy_actor(new_seal.scenario, new_seal.ap_coordinates, template_ap, new_seal.ap_name, 16)
-    editor.copy_actor(new_seal.scenario, new_seal.platform_coordinates, template_platform, new_seal.platform_name, 10)
+    editor.copy_actor(scenario_name, new_seal.ap_coordinates, template_ap, new_seal.ap_name, 16)
+    editor.copy_actor(scenario_name, new_seal.platform_coordinates, template_platform, new_seal.platform_name, 10)
     scenario_file.raw.actors[10][new_seal.platform_name]["components"][0]["arguments"][4]["value"] = new_seal.ap_name
 
-    if new_seal.scenario == "s110_surfaceb":
+    if scenario_name == "s110_surfaceb":
         scenario_map = editor.get_file("gui/minimaps/c10_samus/s000_surface.bmsmsd", Bmsmsd)
     else:
-        scenario_map = editor.get_file(f"gui/minimaps/c10_samus/{new_seal.scenario}.bmsmsd", Bmsmsd)
+        scenario_map = editor.get_file(f"gui/minimaps/c10_samus/{scenario_name}.bmsmsd", Bmsmsd)
 
     scenario_map.raw["tiles"][new_seal.tile_index]["icons"] = ListContainer([CHOZO_SEAL_ICON])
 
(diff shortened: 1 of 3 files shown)


src/open_samus_returns_rando/specific_patches/chozo_seal_patches.py
```diff
@@ -45,5 +45,270 @@ def patch_dna_check(editor: PatcherEditor):
...
+
 def patch_chozo_seals(editor: PatcherEditor):
     patch_dna_check(editor)
+    ensure_dependencies(editor)
+    add_area3b_seal(editor)
+    add_area5_seal(editor)
+    add_area6b_seal(editor)
+    add_area10_1_seal(editor)
+    add_area10_2_seal(editor)
+    add_surfaceb_seal(editor)
```
ThanatosGit, line 314: Please don't matter endless copies of the same function, if all you really want to do is the same just with different parameters.

ThanatosGit (changes requested):

src/open_samus_returns_rando/specific_patches/chozo_seal_patches.py
ThanatosGit: The rewrite is completly missing my point here about `do [...] the same just with different parameters`.  It has swapped endless copies of a function for a new spaghetti function which checks for the scenario every second line.

You already started to use a proper data structure by extracting the coordinates but then sadly stopped.

I was thinking about something like the following. Such that if you want to modify, remove or add another seal, you don't have to touch any code besides adding a `NewChozoSeal` object to the list.

```python
import typing

class NewChozoSeal(typing.NamedTuple):
    scenario: str
    coordinates: tuple[(float, float, float)]
    ap_coordinates: tuple[(float, float, float)]
    tile_index: int
    entity_groups: list[str]
    ap_name: str = "LE_RandoDNA"
    platform_name: str = "LE_Platform_RandoDNA"

new_seals = [
    NewChozoSeal(
        "s033_area3b", [-3800.0, -16800.0, 0.0], [-3800.0, -16800.0, 0.0], 22, ["collision_camera_033", "PostGamma_004"]
    )
]

def add_chozo_seals(editor: PatcherEditor, new_seal: NewChozoSeal):
    template_ap = editor.get_scenario("s000_surface").raw.actors[16]["LE_ChozoUnlockAreaDNA"]
    template_platform = editor.get_scenario("s000_surface").raw.actors[10]["LE_Platform_ChozoUnlockAreaDNA"]
    
    CHOZO_SEAL_ICON = Container({
        "actor_name": "LE_RandoDNA",
        "clear_condition": "",
        "icon": "systemmechdna",
        "icon_priority": 4,
        "coordinates": ListContainer(new_seal.coordinates),
    })
    ...
    rest of the code
    ...
   # e.g. the whole entity group part will just look like this
    for entity_group in new_seal.entity_groups:
        scenario_file.add_actor_to_entity_groups(entity_group, new_ap, True)
        scenario_file.add_actor_to_entity_groups(entity_group, new_platform, True)

def patch_chozo_seals(editor: PatcherEditor):
    patch_dna_check(editor)
    for new_seal in new_seals:
        add_chozo_seals(editor, new_seal)
```

src/open_samus_returns_rando/specific_patches/chozo_seal_patches.py
```diff
@@ -45,5 +47,76 @@ def patch_dna_check(editor: PatcherEditor):
     editor.replace_asset(file_name, systemmechdna)
 
 
+class NewChozoSeal(typing.NamedTuple):
+    scenario: str
+    ap_coordinates: tuple[(float, float, float)]
+    platform_coordinates: tuple[(float, float, float)]
```
ThanatosGit, line 53: Oops, that was wrong on my side! We are actually just using a `list[float]` for both.

ThanatosGit (approved):
Nice. Looks much cleaner now. Only one `if-else-block` required for the surface shenanigans 👍

Just change the type where I did it wrong and go for the merge it.

dyceron:
> Nice. Looks much cleaner now. Only one `if-else-block` required for the surface shenanigans 👍
> 
> Just change the type where I did it wrong and go for the merge it.

I kinda wish I could think of these things, but that will come with time I guess. I wonder if it's worth revisiting other files like this. For later.

PR #264, opened 2024-02-02
Fix Metroid BMSAD

ThanatosGit:
The `PatcherEditor` uses a dict `memory_files` when using `get_file`. Manipulating the same asset via `get_parsed_asset` + `replace_asset` and at the other code part with `get_file` is bad.

Should change the other usages of `get_parsed_asset` later.

### DIFF
--- a/src/open_samus_returns_rando/specific_patches/metroid_patches.py
+++ b/src/open_samus_returns_rando/specific_patches/metroid_patches.py
@@ -18,7 +18,7 @@ def _patch_metroids(editor: PatcherEditor):
     ]
 
     for metroid_file in METROID_FILES:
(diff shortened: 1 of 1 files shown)


dyceron (approved):
Was it overwriting the Metroid changes or just not applying at all?

ThanatosGit:
The PB part uses `get_file`, which means `editor`  reads the asset and holds the changed files in the `memory_files` dict until the very end of the pacher where we call `flush_modified_assets`.
Changes in between, by using directly `get_parsed_asset` and `replace_asset` like in the metroid part before, are overwritten because the `memory_files` cache doesn't see those changes.

dyceron:
Ah ok. That makes sense.

PR #266, opened 2024-02-05
Add DNA seals to all areas

dyceron:
Seals have been added to all remaining areas. The seal in `surfaceb` has been temporarily disabled until we figure out why it breaks the Ridley fight. This means we currently have 9 seals to work with for DNA hints.

### Area 1
image

### Area 2
image

### Area 6
Lower
image

Upper (Moved from Area 8 lower)
image

### Area 7
image


dyceron:
Before this gets merged, I was thinking maybe we should revamp this a bit. Idk if having two Seals in a8 is ideal, nor am I sure if locking hints behind the queen is a good idea.

ThanatosGit (approved):
From the technical standpoint, I can easily approve it because that's exactly the reason why we added such a structure for the seals ;)

From the design standpoint feel free to do what you think balance it nicely.

PR #270, opened 2024-02-11
Display room names in-game

dyceron:
Takes a dict of collision_cameras and the rdv database room names and displays the room name on screen. Used the small text because the long names are cut off.

image

There's only one issue which I haven't figured out is that when a new scenario is loaded, the room name doesn't display until a subarea charge occurs. While annoying, everything else is fully functional. It would be nice to fix this issue, but I'm fine with this otherwise.

Future implementations could figure out how to display this on the top screen and to also support fading out.

(Video doesn't do the text justice. It looks better than is shown)

https://github.com/randovania/open-samus-returns-rando/assets/38679103/472487f9-0dbe-4263-b5ef-b0ecc69ce926


src/open_samus_returns_rando/files/custom/room_names.lua
```diff
@@ -0,0 +1,91 @@
...
+RoomNameGui = RoomNameGui or {
+    ui = nil,
+    cameraDict = RANDO_CC_DICTIONARY,
+}
+
+function RoomNameGui.GetRoomName(camera)
+    local dict = RoomNameGui.cameraDict
+    local scenario = Scenario.CurrentScenarioID
+    if dict == nil then
+        Game.LogWarn(0, "No camera dict present!")
```
ThanatosGit, line 12: We have nothing for logging. I would prefer to save the bytes.

src/open_samus_returns_rando/files/custom/room_names.lua
```diff
@@ -0,0 +1,91 @@
...
+        return
+    end
+
+    local scenario_dict = dict[scenario]
+    if scenario_dict == nil then
+        Game.LogWarn(0, scenario .. " has nil dict!")
```
ThanatosGit, line 18: same

src/open_samus_returns_rando/files/custom/room_names.lua
```diff
@@ -0,0 +1,91 @@
...
+    RoomNameGui.ui = ui
+
+    local current_cc = ""
+
+    RoomNameGui.Update(current_cc)
+end
+
+function RoomNameGui.Update(new_cc)
+    if type(new_cc) ~= "string" then
+        GUI.LogWarn(0, "collision camera is not string")
```
ThanatosGit, line 72: Same but also I don't think that this function exists on `GUI` only on `Game` afaik.

src/open_samus_returns_rando/files/custom/room_names.lua
```diff
@@ -0,0 +1,91 @@
...
+        BlinkColorB = "1.00000",
+        BlinkAlpha = "1.00000",
+        Blink = "-1.00000"
+    })))
+
+    RoomNameGui.ui = ui
+
+    local current_cc = ""
+
+    RoomNameGui.Update(current_cc)
```
ThanatosGit, line 67: What's the purpose of updating it with an empty text? This function always recreates the object anyways, so it should be empty?

src/open_samus_returns_rando/files/custom/room_names.lua
```diff
@@ -0,0 +1,91 @@
...
+    local hud_text = GUI.GetDisplayObject("IngameMenuRoot.IngameMenuComposition.LowerComposition.LowerInfoComposition.RoomNameText")
+    local room_name = RoomNameGui.GetRoomName(new_cc)
+
+    -- This cc isn't in the db so this simulates the room dict
+    if Scenario.CurrentScenarioID == "s000_surface" and new_cc == "collision_camera_017" then
+        room_name = "Transport to Area 8"
+    end
+
+    if room_name == nil then
+        Game.LogWarn(0, string.format("Couldn't find name for %s/%s", "", new_cc))
```
ThanatosGit, line 85: same

src/open_samus_returns_rando/files/custom/room_names.lua
```diff
@@ -0,0 +1,91 @@
...
+        return
+    end
+
+    local hud_text = GUI.GetDisplayObject("IngameMenuRoot.IngameMenuComposition.LowerComposition.LowerInfoComposition.RoomNameText")
+    local room_name = RoomNameGui.GetRoomName(new_cc)
+
+    -- This cc isn't in the db so this simulates the room dict
+    if Scenario.CurrentScenarioID == "s000_surface" and new_cc == "collision_camera_017" then
+        room_name = "Transport to Area 8"
+    end
```
ThanatosGit, line 82: Just add it to the dictionary at the `collision_camera_table.py`

src/open_samus_returns_rando/files/custom/room_names.lua
```diff
@@ -0,0 +1,91 @@
...
+    RoomNameGui.Update(current_cc)
+end
+
+function RoomNameGui.Update(new_cc)
+    if type(new_cc) ~= "string" then
+        GUI.LogWarn(0, "collision camera is not string")
+        return
+    end
+
+    local hud_text = GUI.GetDisplayObject("IngameMenuRoot.IngameMenuComposition.LowerComposition.LowerInfoComposition.RoomNameText")
```
ThanatosGit, line 76: Isn't this just `RoomName.ui` ?

ThanatosGit (changes requested):
If this is yet another feature for stable, I would prefer that the ILS thing is checked. I have no idea how the overhead of two additional files is. Especially when one of those files is a 300+ lines dict where the compiled code will be nearly the same size as the non-compiled ascii version.

src/open_samus_returns_rando/files/custom/room_names.lua
dyceron: In Dread this was the function to update the current cc. In MSR, said function doesn't exist and I can't seem to find a replacement, hence why the first cc that you load into in a scenario doesn't show the name until after a sub area change occurs.

src/open_samus_returns_rando/files/custom/room_names.lua
dyceron: I don't understand. I did this to avoid having to create a placeholder room in the database. How would I add to the python file directly?

src/open_samus_returns_rando/files/custom/room_names.lua
dyceron: Yeah probably

dyceron: Didn't seem to work when I tried changing it so leaving it alone.

ThanatosGit (approved):

PR #273, opened 2024-02-16
Use super missile model with red texture for missile launcher

ThanatosGit:
https://github.com/randovania/open-samus-returns-rando/assets/117127188/d13bb755-9b65-4d3e-8c6c-1f8fa948e6f1


src/open_samus_returns_rando/pickups/model_data.py
```diff
@@ -285,16 +285,10 @@ class ModelData:
         ),
     ),
     "powerup_missilelauncher": ModelData(
-        bcmdl_path="actors/weapons/missile/models/missile.bcmdl",
+        bcmdl_path="actors/items/powerup_randomissile/models/powerup_randomissile.bcmdl",
```
dyceron, line 288: While I understand why you named it this way, I think it'd be better to follow the naming scheme/style of vanilla, especially as we add more textures/models for custom stuff. It's more of a nitpick, but more consistent.

ThanatosGit: I never looked up how a bcmdl file works. It has some weird lookup stuff string in it? That's why I just quickly used the first name I had in mind which has the same length as `powerup_supermissile`. Same for the texture.

ThanatosGit: The length definitely matters because the textures are referenced by their name. If you change the length you destroy every offset in the file. In odr there are also many files stretched with `___` etc. for that reason.

But I think, at least in msr, it uses the name of the texture defined in the bctex and not the name of the file itself. Can play around with it.
What would be your preferred name for the bcmdl and bctex?
`powerup_missile`? (like for supers...just without the super)

dyceron: I'd like to try to be consistent across the board, so I'd do `missilelauncher` since the pickup is named as such in the patcher and rdv.

ThanatosGit: Scratch what I was saying before: The bcmdl reference to the texture just searches for the file name. As I have no idea how BCMDL in SR are built, I need to keep the length of the texture name otherwise I break it.

dyceron (approved):

PR #275, opened 2024-02-18
Prevent `HP_Item_001` from being deleted

dyceron:
In area2b, if you lay a Power Bomb next the door in `collision_camera009`, the radius of the explosion reaches and breaks the Bomb blocks in `collision_camera012`. If you enter CC12 while the PB is still active, the blocks will be broken and the pickup is visible and can be collected. If you enter CC12 after the PB explosion is gone, the blocks will have been broken, but the block containing `HP_Item_001` despawns, causing the pickup to be lost forever.

This fix avoids this bug by making the pickup freestanding, and then creating a custom Bomb block in place of the pickup block which does not respawn after being broken. In this way, the vanilla behavior is simulated and the pickup is always accessible even if a PB is laid in CC9. The only side effect of this is that the item no longer is tied to the block, but under normal circumstances this is perfectly acceptable and prevents ruining a seed.

### DIFF
--- a/src/open_samus_returns_rando/specific_patches/static_fixes.py
+++ b/src/open_samus_returns_rando/specific_patches/static_fixes.py
@@ -1,3 +1,4 @@
+from construct import Container, ListContainer
 from mercury_engine_data_structures.formats import Bmsad, Bmsbk, Bmtun
 from open_samus_returns_rando.patcher_editor import PatcherEditor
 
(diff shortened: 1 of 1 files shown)


ThanatosGit (commented):
I never looked how the hidden pickups or blocks works.

You empty a string on one of the actor's component: I guess that's the block the item is hidden behind?
But isn't there already a block? Because this patch is only adding a block and not removing one.  I would have expected to see one block being removed and one being added or just change the existing one.

Is a block or `sg_casca38` associated with any cc?

dyceron:
Yes the component I empty is the block that the item resides in. This can be changed to any breakable block type or nothing (freestanding). The block only exists because of that value. This applies to all hidden items and why the items arent accessible until the block is broken, even if you clip inside it. So yeah I am removing a block and adding one. The removed one is tied to the actor and not the area file.

The actor is only part of the actor group in cc12. `sg_casca38` i assume is the texture for `vignette_casca38.bcmdl`.

ThanatosGit (approved):

ThanatosGit:
.. and gj, dcr. I would never thought about that this has something to do with pbs in another area. I wonder why the blocks react to it if they are not part of the cc and shouldn't be loaded.

dyceron:
I think it's because in this one spot by the door the cc's overlap, so the block is probably loaded but from the wrong cc so it just screws things up.

Issue #276, opened 2024-02-18
Missile Reserve Tanks don't refill Missiles


dyceron:
For some reason only Super Missiles get refilled, but never Missiles. Has this always been broken?

ThanatosGit:
Uh, no idea. I think it was not broken?
Maybe because you removed the `_MAX` thing from the gun thing to make randomized missile launcher?
(My wording is on top in the early morning)

dyceron:
Yep that's exactly why. Well that's great.

ThanatosGit:
Maybe we can do both by using a different string at the gun? Like `ITEM_,BLA` and always keep the value to the same as the max. for missiles.

dyceron:
I tried setting it to `_CURRENT` and I'm back to not being able to select Supers without Missiles again.

dyceron:
We could slightly mitigate this for seeds that start with Missile Launcher. Set the gun string back to vanilla if starting with it, leave it as `""` if not. Obviously, this doesn't solve our issue, but it does bring back some functionality at least. I'm always gonna prioritize shuffled Missiles over the Reserve Tank. Could always slap a "doesnt work if shuffled Missiles" if we go this route into rdv (I want to make a known issues tab anyway so this would be perfect for that if we can't figure it out).

dyceron:
Something like this (though much cleaner than this of course) 
![image](https://github.com/randovania/open-samus-returns-rando/assets/38679103/e6bcac58-a575-48d1-919f-d296a968d963)

dyceron:
I doubt we can modify bmsad's during runtime, so that's out of the question.

dyceron:
Possible relevant function
![image](https://github.com/randovania/open-samus-returns-rando/assets/38679103/38d15308-065b-40df-8f4c-a59f69632065)

dyceron:
From the Discord

> I kinda have a rough workaround for this but its flawed atm
> I added a new item ID upon file creation which value is set to the missile_tank_size config. The gun component for Missiles has also been changed to use this new ID. Issue is that since its set to a certain value, if you have less Missiles than the config, upon activation of the reserve, your missile count temporarily increases to match, which is the bad part. 
> If there's any way to have it only refill what you have while still keeping the max at the config value then we could get around this issue
> aka vanilla behavior but with shuffled launcher basically

[...]

PR #277, opened 2024-02-18
Fix Missile Reserve Tanks not refilling Missiles

dyceron:
- Move Launcher Shuffle to `lua_editor`
- Fix default `missiles_per_tank` count
- ~~Disable gun component only if shuffled launcher~~

~~The bug in #276 still exists with this fix, but this at least makes the Missile Reserve Tank fully usable if starting with Missile Launcher. Dunno if we should try to fully fix this in this PR to make it work with shuffled launcher, or just throw a warning in RDV that this bug exists and merge this one for now.~~

### DIFF
--- a/src/open_samus_returns_rando/lua_editor.py
+++ b/src/open_samus_returns_rando/lua_editor.py
@@ -215,7 +215,7 @@ class LuaEditor:
         if "ITEM_WEAPON_MISSILE_LAUNCHER" in inventory and "ITEM_MISSILE_TANKS" in inventory:
             inventory["ITEM_WEAPON_MISSILE_MAX"] = inventory.pop("ITEM_MISSILE_TANKS")
         else:
-            # For the gun component to work without launcher, any value is sufficient
+            # For the gun component to work without launcher, any value > 0 is sufficient
             inventory["ITEM_MISSILE_CHECK"] = 1
             samus_bmsad = editor.get_file(
                 "actors/characters/samus/charclasses/samus.bmsad", Bmsad
--- a/src/open_samus_returns_rando/lua_editor.py
+++ b/src/open_samus_returns_rando/lua_editor.py
@@ -1,6 +1,7 @@
 import itertools
 
 from construct import Container
+from mercury_engine_data_structures.formats import Bmsad
 from mercury_engine_data_structures.formats.lua import Lua
 
 from open_samus_returns_rando.constants import ALL_SCENARIOS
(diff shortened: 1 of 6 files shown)


dyceron:
With the latest commit, max refill value is now tied to collecting the launcher and how many missiles it gives + all tanks collected before the getting the launcher. It caps out at the value set in the config, so no need to worry about that end. All we need now is to add the tanks collected after the launcher and we can have both launcher shuffled and missile reserves working.

dyceron:
Update: The Missile Reserve Tank now properly supports refilling Missiles regardless if launcher is shuffled. If launcher isn't shuffled, the gun component just uses the original value of `ITEM_WEAPON_MISSILE_MAX`. If it is shuffled, it uses a new item ID which updates to match the value of `ITEM_WEAPON_MISSILE_MAX` (the min refill value). The max refill value is the value of `missile_tank_size` set in the configuration.

src/open_samus_returns_rando/lua_editor.py
```diff
@@ -213,6 +214,13 @@ def _create_custom_init(self, editor: PatcherEditor, configuration: dict) -> str
         # use _MAX if main is unlocked to unlock the ammo too
         if "ITEM_WEAPON_MISSILE_LAUNCHER" in inventory and "ITEM_MISSILE_TANKS" in inventory:
             inventory["ITEM_WEAPON_MISSILE_MAX"] = inventory.pop("ITEM_MISSILE_TANKS")
+        else:
+            # For the gun component to work without launcher, any value > 0 is sufficient
+            inventory["ITEM_MISSILE_CHECK"] = 1
+            samus_bmsad = editor.get_file(
+                "actors/characters/samus/charclasses/samus.bmsad", Bmsad
+            )
+            samus_bmsad.raw["components"]["GUN"]["functions"][20]["params"]["Param5"]["value"] = "ITEM_MISSILE_CHECK"
```
ThanatosGit, line 223: Can't we just move the whole code back to where it was and always use `ITEM_MISSILE_CHECK` ?

src/open_samus_returns_rando/lua_editor.py
```diff
@@ -213,6 +214,13 @@ def _create_custom_init(self, editor: PatcherEditor, configuration: dict) -> str
         # use _MAX if main is unlocked to unlock the ammo too
         if "ITEM_WEAPON_MISSILE_LAUNCHER" in inventory and "ITEM_MISSILE_TANKS" in inventory:
             inventory["ITEM_WEAPON_MISSILE_MAX"] = inventory.pop("ITEM_MISSILE_TANKS")
+        else:
+            # For the gun component to work without launcher, any value > 0 is sufficient
+            inventory["ITEM_MISSILE_CHECK"] = 1
```
ThanatosGit, line 219: You are missing the case where you start with the Missile Launcher + Ammo but haven't collected a tank in-between.
Like in the oter comment: I would suggest to just always use `ITEM_MISSILE_CHECK` and use 
```
"ITEM_MISSILE_CHECK: max(1, inventory.get("ITEM_WEAPON_MISSILE_MAX, 0))
``` 
in the `final_inventory` here.

ThanatosGit (changes requested):
This is very much what I suggested at the very beginning https://github.com/randovania/open-samus-returns-rando/issues/276#issuecomment-1950976099
I was somehow under the impression that you already said that this doesn't work oO

dyceron:
> This is very much what I suggested at the very beginning [#276 (comment)](https://github.com/randovania/open-samus-returns-rando/issues/276#issuecomment-1950976099) I was somehow under the impression that you already said that this doesn't work oO

Nope I didn't try this specifically. My test before was flawed and not very good.

src/open_samus_returns_rando/lua_editor.py
dyceron: Hmm I guess I figured it'd be better to only have this be different if not shuffled. At one point I did check for the tunable value, but I removed that so there's no real benefit anymore. I can revert it back.

src/open_samus_returns_rando/lua_editor.py
dyceron: Im not sure what you mean by this. If you start with launcher this else block is ignored.

ThanatosGit: Whoops yeah. Got confused. The mentoned thing would only be valid if we just always use `ITEM_MISSILE_CHECK`.

ThanatosGit (approved):

PR #280, opened 2024-02-21
Only enable Missile Reserve Tank if Missile Launcher is in inventory

### LINKED ISSUE #253: Missile Reserve Tank activates on scenario load and unpausing
If you shuffle Missile Launcher and have a Missile Reserve Tank which is full, everytime you load a scenario it immediately activates. It probably detects that `ITEM_WEAPON_MISSILE_CURRENT` is 0, and thus activates. This is annoying since it's a waste of a Reserve, time, and requires refilling in the current scenario if you want to use it.

dyceron:
Given the current implementation of shuffled Missile Launcher, this is the simplest solution to preventing the Missile Reserve Tank from activating on its own without Missiles. Now, it checks for Missile Launcher in the inventory to activate. The only downside is that having only Super Launcher means you can't utilize the Reserve Tank to your advantage. Maybe in the future we can have the Reserve Tank be independent of the Launcher, but this is the best option I'm willing to take for now.

Fixes #253

### DIFF
--- a/src/open_samus_returns_rando/files/templates/custom_init.lua
+++ b/src/open_samus_returns_rando/files/templates/custom_init.lua
@@ -24,7 +24,7 @@ function Init.InitGameBlackboard()
       Blackboard.SetProp("PLAYER_INVENTORY", "ITEM_ADN", "f", current_amount + 1)
     end
     if string.sub(_FORV_3_, 1, 17) == "ITEM_RESERVE_TANK" then
-      if (_FORV_3_ == "ITEM_RESERVE_TANK_MISSILE" and Blackboard.GetProp("PLAYER_INVENTORY", "ITEM_WEAPON_MISSILE_LAUNCHER")) or _FORV_3_ ~= "ITEM_RESERVE_TANK_MISSILE" then
+      if _FORV_3_ ~= "ITEM_RESERVE_TANK_MISSILE" or Blackboard.GetProp("PLAYER_INVENTORY", "ITEM_WEAPON_MISSILE_LAUNCHER") then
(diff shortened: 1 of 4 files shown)


src/open_samus_returns_rando/files/templates/custom_init.lua
```diff
@@ -24,8 +24,13 @@ function Init.InitGameBlackboard()
...
     if string.sub(_FORV_3_, 1, 17) == "ITEM_RESERVE_TANK" then
-      Blackboard.SetProp("GAME", _FORV_3_ .. "_ACTIVE", "b", true)
-      Blackboard.SetProp("GAME", _FORV_3_ .. "_FULL", "b", true)
+      if _FORV_3_ == "ITEM_RESERVE_TANK_MISSILE" and Blackboard.GetProp("PLAYER_INVENTORY", "ITEM_WEAPON_MISSILE_LAUNCHER") then
+        Blackboard.SetProp("GAME", "ITEM_RESERVE_TANK_MISSILE_ACTIVE", "b", true)
+        Blackboard.SetProp("GAME", "ITEM_RESERVE_TANK_MISSILE_FULL", "b", true)
+      elseif _FORV_3_ ~= "ITEM_RESERVE_TANK_MISSILE" then
+        Blackboard.SetProp("GAME", _FORV_3_ .. "_ACTIVE", "b", true)
+        Blackboard.SetProp("GAME", _FORV_3_ .. "_FULL", "b", true)
+      end
```
ThanatosGit, line 33: Both flows doing the exact same thing. Can just be combined in one if-clause with an `or`.

ThanatosGit (approved):

PR #287, opened 2024-02-22
Fix s036_area3c Gammas Crash

### LINKED ISSUE #286: Crash entering s036_area3c cc14 because of Gamma_007 spawngroup
Entering cc11 after killing the Gamma without going in the right cc first causes a memory issue when attempting to enter cc14

https://clips.twitch.tv/CoyEvilSnakeBuddhaBar-qO_WL0JsEEpcPIui

dyceron:
Delaying or removing the CC changes for Gamma_006 until after Gamma_007 is defeated avoids any conflicts and removes the crash/memory errors. I chose the former option since it has the most accurate results in game.

Fixes #286

### DIFF
--- a/src/open_samus_returns_rando/files/levels/s036_area3c.lua
+++ b/src/open_samus_returns_rando/files/levels/s036_area3c.lua
@@ -322,13 +322,7 @@ function s036_area3c.OnGamma_007_B_Generated(_ARG_0_, _ARG_1_)
   end
 end
 function s036_area3c.OnEnter_Gamma_007_Dead()
-  if Scenario.ReadFromBlackboard("Arena_Gamma_007_AllDead", false) then
-    Game.SetSubAreaCurrentSetup("collision_camera_013", "PostGamma_007", true)
-    Game.SetSubAreaCurrentSetup("collision_camera_027", "PostGamma_007", true)
-    if Game.GetEntity("SpawnGroup008") ~= nil then
-      Game.GetEntity("SpawnGroup008").SPAWNGROUP:EnableSpawnGroup()
-    end
-  end
+  s036_area3c.OnSubAreaChange()
 end
 function s036_area3c.OnGenericCrawler_Generated(_ARG_0_, _ARG_1_)
   if _ARG_1_ ~= nil then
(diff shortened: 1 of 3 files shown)


src/open_samus_returns_rando/files/levels/s036_area3c.lua
```diff
@@ -322,13 +322,7 @@ function s036_area3c.OnGamma_007_B_Generated(_ARG_0_, _ARG_1_)
...
 end
 function s036_area3c.OnEnter_Gamma_007_Dead()
-  if Scenario.ReadFromBlackboard("Arena_Gamma_007_AllDead", false) then
-    Game.SetSubAreaCurrentSetup("collision_camera_013", "PostGamma_007", true)
-    Game.SetSubAreaCurrentSetup("collision_camera_027", "PostGamma_007", true)
-    if Game.GetEntity("SpawnGroup008") ~= nil then
-      Game.GetEntity("SpawnGroup008").SPAWNGROUP:EnableSpawnGroup()
-    end
-  end
+  s036_area3c.OnSubAreaChange()
```
ThanatosGit, line 325: What's that supposed to do? It's missing every parameter for this function.

dyceron: It's supposed to call that function, but maybe it's breaking because of the missing parameters. Or maybe it just won't work like this at all (though why does this not crash console but it does crash citra)

ThanatosGit: I was aware of what it is doing technically 😛. It was more about what this logically does.

dyceron: Oh. The goal is to only change the Gamma_007 ccs after entering the transport cc. For some reason if enter the side room cc first then leave the Gamma section the game will crash if you enter the right cc past the transport. I didn't test leaving the scenario or taking the upper tunnel but yeah it breaks otherwise on console. This change makes it work on console, but crashes on citra, which is why I'm frustrated and confused

ThanatosGit:
@dyceron can you test this? It's not calling the OnDead anymore after the metroid dies but both rooms (left, right) from the metroid section  have a vanilla trigger to call it.
I have no idea if this breaking other stuff! Might even be able to not doing the teleport thing anymore for the other problematic part

ThanatosGit:
...and the starter_preset should be reverted if this works on console.

dyceron:
Yes I can test it layer. If this works could this be applied to all of the multi room Gammas?

ThanatosGit:
I never checked if the radar actually activates again later :D 
Probably need to tune that change

ThanatosGit (approved):

PR #288, opened 2024-02-24
Change a3 gamma crumble to not respawn

dyceron:
Prevents a possible softlock without Bombs/enough PBs while still leaving the Gamma event accessible

https://github.com/randovania/open-samus-returns-rando/assets/38679103/6c0ab3ef-0d4c-4fef-ac2b-90343522cf16


ThanatosGit (approved):
Hmmm...isn't it kind of the fun that you are required to have something something bomb to access this metroid without softlocking?

PR #289, opened 2024-02-24
Proper fix multis

ThanatosGit:
Reverts https://github.com/randovania/open-samus-returns-rando/pull/287

Makes a change to the SG and SP such that disabling the SG is enough to deactivate the radar.
Activates the TG for the Intro of 007 by default.
Needs to deactivate the trigger as part of the callback to not let the metroid respawn while the metroid setup is still active.

Does this the whole stuff work for 3DS?  Also regarding the whole crash when going to the right of teleporter room? If so, we can work with it further to rework all other multi gammas and get rid of the stupid warping stuff.
Also be aware that disabling the Trigger is just hard coded now...so killing 006 first will deactivate 007.

### DIFF
--- a/src/open_samus_returns_rando/files/custom/scenario.lua
+++ b/src/open_samus_returns_rando/files/custom/scenario.lua
@@ -45,7 +45,6 @@ end
 function Scenario.SetMetroidSpawngroupOnCurrentScenario(created_actor, group_name, is_multi)
   if created_actor ~= nil and created_actor.sName ~= nil then
     CurrentScenario.currentMetroidSpawngroup = group_name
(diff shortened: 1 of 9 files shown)


dyceron:
Latest commit has all multi-room Gamma working as expected in vanilla. Things to still look into:
- Some Gammas spawn from their valves instead of the typical spawn point
- Reloading a checkpoint also re-enables the trigger for the Gamma, so they can be fought again
  - Checkpoints for them are still disabled to prevent this
- Cleanup any unused/unnecessary code that might have been left behind

dyceron (approved):

PR #290, opened 2024-02-25
Prevent `s067_area6c: HiddenPowerup002` from being deleted with PB

dyceron:
A5 has another hidden item that is similar to A2 which can be destroyed from another CC. Since accessing the pickup requires going through a couple CCs, the CC which the actor resides will be reloaded, so no need to add a new block. Just ensure the actor exists in both CCs.

### DIFF
--- a/src/open_samus_returns_rando/specific_patches/static_fixes.py
+++ b/src/open_samus_returns_rando/specific_patches/static_fixes.py
@@ -161,8 +161,8 @@ def increase_pb_drop_chance(editor: PatcherEditor):
             drop["fPowerBombProbability"]["value"] *= 2
 
 
-def fix_area2b_hp_item_001_deletion(editor: PatcherEditor):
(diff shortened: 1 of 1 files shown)


ThanatosGit:
I don't understand the explanation. 
Can't you just destroy it from 005 and go to 007 directly? 
How is that different from the one before where you destroy it from 009 and go directly to 012?

dyceron:
> I don't understand the explanation. Can't you just destroy it from 005 and go to 007 directly? How is that different from the one before where you destroy it from 009 and go directly to 012?


Roughly how this works

A2
- Lay pb in 009 -> 012 (visual bug/deleted)
OR
- Lay pb in 009 -> 041 (small access room, 012 is reloaded) -> 012

A5
- Lay pb in 005 -> 007 top -> 023 (teleporter cc, 007 bottom is reloaded) -> 007 bottom

ThanatosGit (approved):

PR #294, opened 2024-03-02
Ensure asset dependencies are added to the correct pkgs

dyceron:
Just like with shuffled pickups, we have to ensure the new Chozo Seal assets are added to the correct pkgs, otherwise Diggernaut breaks and the fight doesn't end.

### DIFF
--- a/src/open_samus_returns_rando/specific_patches/chozo_seal_patches.py
+++ b/src/open_samus_returns_rando/specific_patches/chozo_seal_patches.py
@@ -2,7 +2,6 @@ import typing
 
 from construct import Container, ListContainer
 from mercury_engine_data_structures.formats import Bmsad, Bmsmsd
-from open_samus_returns_rando.constants import get_package_name
 from open_samus_returns_rando.patcher_editor import PatcherEditor
 
 SCRIPT_COMPONENT = Container({
@@ -166,9 +165,8 @@ def update_item_seals(editor:PatcherEditor):
             scenario.raw.actors[16][seal]["type"] = "chozoseal"
 
         # Dependencies
-        for level_pkg in editor.get_level_pkgs(scenario_name):
-            for asset in editor.get_asset_names_in_folder("actors/props/chozoseal"):
-                editor.ensure_present(get_package_name(level_pkg, asset), asset)
+        for asset in editor.get_asset_names_in_folder("actors/props/chozoseal"):
+            editor.ensure_present_in_scenario(scenario_name, asset)
 
 
 def patch_chozo_seals(editor: PatcherEditor):
--- a/src/open_samus_returns_rando/specific_patches/chozo_seal_patches.py
+++ b/src/open_samus_returns_rando/specific_patches/chozo_seal_patches.py
@@ -166,7 +166,7 @@ def update_item_seals(editor:PatcherEditor):
             scenario.raw.actors[16][seal]["type"] = "chozoseal"
 
         # Dependencies
-        for level_pkg in editor.get_all_level_pkgs():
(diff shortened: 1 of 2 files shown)


src/open_samus_returns_rando/specific_patches/chozo_seal_patches.py
```diff
@@ -165,8 +166,9 @@ def update_item_seals(editor:PatcherEditor):
             scenario.raw.actors[16][seal]["type"] = "chozoseal"
 
         # Dependencies
-        for asset in editor.get_asset_names_in_folder("actors/props/chozoseal"):
-            editor.ensure_present_in_scenario(scenario_name, asset)
+        for level_pkg in editor.get_all_level_pkgs():
```
ThanatosGit, line 169: Why all level pkgs? This should only be relevant for the scenarios where we add stuff.

ThanatosGit, line 169: This is also iterating over all level pkgs for each chozo seal which is definitely unnecessary.

src/open_samus_returns_rando/specific_patches/chozo_seal_patches.py
```diff
@@ -165,8 +166,9 @@ def update_item_seals(editor:PatcherEditor):
...
+            for asset in editor.get_asset_names_in_folder("actors/props/chozoseal"):
+                editor.ensure_present(get_package_name(level_pkg, asset), asset)
```
ThanatosGit, line 171: The thing this PR is changing, is that it respects the vanilla distribution of files between the normal pkg and `_discardables.pkg`.

If that's the issue here, it should be fixed in a more common way or, at least, should also take care of the other code parts where it is done wrongly. E.g. the above method `add_chozo_seals` does the same and is just in sight here. I'm very sure that there are more parts.

A more common way would be preferred because I assume that this was also the culprit with Ridley.

ThanatosGit (changes requested):

src/open_samus_returns_rando/specific_patches/chozo_seal_patches.py
dyceron: Should I list out all scenarios where we have the custom assets, or just a6 since that's where it breaks things?

ThanatosGit: No, just look what was the code before. It was already only adding the dependencies to the scenario and it was also only using the scenario name of the currently processed chozo seal in the loop

dyceron: So this would work?
```suggestion
        for asset in editor.get_asset_names_in_folder("actors/props/chozoseal"):
            editor.ensure_present(get_package_name(scenario_name, asset), asset)
```

src/open_samus_returns_rando/specific_patches/chozo_seal_patches.py
dyceron: Should we change all instances of `ensure_present_in_scenario` to `ensure_present`?

What do you mean by common way?

ThanatosGit: From my point of view, you have shown that we have a bug in `ensure_present_in_scenario` because it allows you to add a file to the wrong package.
I think the best would be to actually revert the changes here and just add the `get_package_name` call in `ensure_present_in_scenario`, so it can fixed in all places?!

I'm curious if ridley will work then with a chozo seal in surfaceb

ThanatosGit (approved):

PR #297, opened 2024-03-10
Door lock

ThanatosGit:
I have no idea what other issues we had but could still be fixed in a following PR.
(also i don't like that our old friend from dread (two missiles for a shield) are back)

### DIFF
--- a/src/open_samus_returns_rando/specific_patches/door_patches.py
+++ b/src/open_samus_returns_rando/specific_patches/door_patches.py
@@ -218,7 +218,7 @@ class DoorType(Enum):
         "actors/props/doorchargecharge", "actors/props/door/fx", "sounds/props/doorchargecharge"
     ])
     WAVE_BEAM = ("wave_beam", ActorData.DOOR_POWER, True, ActorData.SHIELD_WAVE_BEAM, [
-        "actors/props/doorwave", "sounds/props/doorwave"
+        "actors/props/doorwave", "sounds/props/doorwave", "system/fx/textures/blood_gray.bctex",
     ])
     SPAZER_BEAM = ("spazer_beam", ActorData.DOOR_POWER, True, ActorData.SHIELD_SPAZER_BEAM, [
         "actors/props/doorspazerbeam", "sounds/props/spazerdoor"
--- a/src/open_samus_returns_rando/specific_patches/door_patches.py
+++ b/src/open_samus_returns_rando/specific_patches/door_patches.py
@@ -1,8 +1,9 @@
 
 import copy
+from enum import Enum
 
 from construct import Container, ListContainer
-from mercury_engine_data_structures.formats import Bmsad, Lua
+from mercury_engine_data_structures.formats import Bmsad, Bmsld, Lua
 from open_samus_returns_rando.files import files_path
 from open_samus_returns_rando.patcher_editor import PatcherEditor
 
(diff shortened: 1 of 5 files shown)


ThanatosGit:
Oh, our settings needs to be updated

src/open_samus_returns_rando/specific_patches/door_patches.py
```diff
@@ -187,9 +188,153 @@ def _patch_one_way_doors(editor: PatcherEditor):
...
+                    ActorData.SHIELD_PLASMA_BEAM, ["actors/props/doorcreature"])
+    MISSILE = ("missile", ActorData.DOOR_POWER, True, ActorData.SHIELD_MISSILE, [
+        "actors/props/doorshield", "actors/props/doorshieldmissile"
+    ])
+    SUPER_MISSILE = ("super_missile", ActorData.DOOR_POWER, True, ActorData.SHIELD_SUPER_MISSILE, [
+        "actors/props/doorshield", "actors/props/doorshieldsupermissile"
+    ])
+    POWER_BOMB = ("power_bomb", ActorData.DOOR_POWER, True, ActorData.SHIELD_POWER_BOMB, [
+        "actors/props/doorshield", "actors/props/doorshieldpowerbomb"
+    ])
```
dyceron, line 234: I assume with the proper dependency assurance, the new shields play their expected sounds?

ThanatosGit: That would require me to unmute the game 😮.
Added sound dependencies.
Do SM and PB doors actually make any sound? I mean, I see that there is a file for it, i played the file on my pc by opening the bcwav but if I go to a SM or PB door in the game (without any door patches!) and can't hear anything.
Or is this something since "ever" with the rando / duplicated shields?

dyceron: Yes they make sounds :p

ThanatosGit: What about the second part? `Or is this something since "ever" with the rando / duplicated shields?` 
Because I stay right in front of this door and there is nothing 
image

dyceron: I've definitely heard sounds play on the rotated doors. I'm not sure why it wouldn't play at times. Too many sounds maybe? Rotation shouldn't matter since I always hear the Missile shields just fine.

ThanatosGit:
Wave door is missing something something when destroyed:

https://github.com/randovania/open-samus-returns-rando/assets/117127188/64fbb8b3-d1fd-4def-9227-53fdcdb5cb99



May want to exclude this door from being a normal door? 
image

dyceron:
> Wave door is missing something something when destroyed:
> 
I did notice that a while ago. Some missing dependency I guess.
> 
> May want to exclude this door from being a normal door? image

 Any reason for this?

dyceron:
This PB door isn't making a noise, but it could just be a sound overload? I know that's an issue in the Area 5 Chozo Seal normally, but it also could be something else at play here.
image

ThanatosGit:
Ngl, I won't care anymore about sound stuff in this PR.
Every missile shield (i have seen so far) makes noise but for pb and sm it's random for me but this already happens without any door changes. 
So, this will fall into what I said in the opening comment: Could still be fixed in a later PR.

ThanatosGit:
> > May want to exclude this door from being a normal door? image
> 
> Any reason for this?

Because I don't like that it's removing all requirements for this pickup. It get's changed to doorpowerpower which removes the whole fun about getting this pickup.

dyceron:
> > > May want to exclude this door from being a normal door? image
> > 
> > 
> > Any reason for this?
> 
> Because I don't like that it's removing all requirements for this pickup. It get's changed to doorpowerpower which removes the whole fun about getting this pickup.

That's totally fair. I'm ok with excluding this one. All the others are basically in comparison so this works.

ThanatosGit:
I would like to finish the stuff I started before focusing on multiworld.
So, I would just exlude the entry door to diggernaut  in the rdv branch for #303 #304 
Then I would work on #300 and this should be good to go.

dyceron:
> I would like to finish the stuff I started before focusing on multiworld. So, I would just exlude the entry door to diggernaut in the rdv branch for #303 #304 Then I would work on #300 and this should be good to go.

Sounds good to me

ThanatosGit:
Should be good now.

Well, the shield is still added to the left side of DoorManicMinerBot but the door itself is always a doorclosed[power, charge], both shield sides are removed if one is destroyed (as always) and the minimap always shows the closed icon on the left side.

dyceron:
The shield being on both sides for Diggernaut is fine since you'd never see that side anyway.

dyceron (changes requested):
There's only one minor minimap issue still. Any door that is a one-way locked door that hasn't been changed deletes its map icon if either side of the door is opened. Need to change the `clear_condition` of the tile to be empty so it updates properly.

image

ThanatosGit:
> a one-way locked door that hasn't been changed
> 

Wdym? 
If the door is excluded then neither the actor nor the minimal is touched, so I don't see why it should have issues.
If the door is included, then all doors are changed to two way doors, so it has been changed.

dyceron:
This isn't an excluded door. It's a door that was a one way locked door, but now changed into power. The icon is in the map before it is opened, then the icon is removed entirely afterwards. This should be changed to keep the power door icon on the map after it has been opened, which is tied to the clear condition

ThanatosGit:
Ah, so it was supposed to be `has been changed` instead of `hasn't been changed` in the original message?

dyceron:
Oh whoops I definitely misspoke there. Yes it should've been one ways that have been changed

ThanatosGit:
- Removed the clear_condition thing.
- Added missing dependency for plasma doors
- Fixed issue with Wave and Plasma not properly update after being destroyed

dyceron (approved):

PR #309, opened 2024-03-29
Fix for hidden powerups

### LINKED ISSUE #308: More fun with disappearing items due to PBs (Area 5 - Phase Drift Trial West)
Should be fixed before stable release in RDV.
See: https://discord.com/channels/914291389293027329/1222038140823539875

ThanatosGit:
The tile which hides the HiddenPowerup003 was always visible if destroyed from adjacent room and I had no idea how to fix it.

Then I decided to not add the items to any other cc group but just do the "freestanding + place block on top of it" approach for all items.

Haven't looked at all items yet but I leave it here for now.

Fixes #308

### DIFF
--- a/src/open_samus_returns_rando/specific_patches/static_fixes.py
+++ b/src/open_samus_returns_rando/specific_patches/static_fixes.py
@@ -1,3 +1,5 @@
+import copy
+
 from construct import Container, ListContainer
 from mercury_engine_data_structures.formats import Bmsad, Bmsbk, Bmtun
 from open_samus_returns_rando.patcher_editor import PatcherEditor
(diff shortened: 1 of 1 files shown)


dyceron:
This looks fine to me. I was hesitant to do it this way for all the blocks because the one in s067 cc7 has a logical clip to get it without Screw Attack, and now it technically can be grabbed without Bombs if you are standing on top of it. Could always prevent this by making the hitbox of that actor smaller so it can't be grabbed without Bombs.

dyceron:
Otherwise, these should be all of the problematic blocks. I checked over the whole map and these were the only ones that were affected from the wrong cc. We could fix the cheese I mentioned, but pr looks good to go whenever.

ThanatosGit:
Do you had a look at all patched items here? I think, I only tested like 2 or 3 that's why I don't know if everything is fine.

dyceron:
I haven't had the time to test these, but I can later. If it's good I'll merge the pr.

dyceron:
So I tested all 5 of these locations and had no issues, so it seems fine (I couldn't replicate the wrong room thing to begin with). The only issue like I've mentioned is some blocks can be grabbed from above due to the item position and the size of the item, but I guess that's something we'll have to deal with. Also, if the model/texture/fx are large then the item can be barely seen protruding from the block (Aeion Abilities for example). This isn't necessarily a bad thing, but not a huge deal either.

dyceron (approved):
Looks good to me. If we want to do more testing/the minor adjustments to fix the less important stuff then we can, otherwise feel free to merge this.

PR #315, opened 2024-04-06
Change all language files

Miepee:


dyceron:
I think this is probably fine? I'd like @ThanatosGit opinion first. I was mainly unsure about changing `mse_english.txt` since that's where the credits are stored, but maybe it's fine.

ThanatosGit:
Changing it is fine, even though I think it's unecessary. Don't know which region / country uses `mse_english`. Could be english speaking without being in europe or in america. But afaik australia doesn't use it, guess they use `eu_english` because they have PAL there anyways?

Whatever, shouldn't do any harm and afaik credits are always pulled from `mse_english`, so it's not neeed to touch `credits.py` https://github.com/randovania/open-samus-returns-rando/blob/77cb6c24eb398e1418f2298e3fbae2d0a7d0aa3e/src/open_samus_returns_rando/misc_patches/credits.py#L70-L100

ThanatosGit (approved):

PR #316, opened 2024-04-06
Remote connector

ThanatosGit:
Stuff to fix:
- [x] ~~Suit lua is too large for the 4096 bytes long buffer. Yikes!~~  `'` was not escaped. Suit Lua is 3936 bytes long, so it still fits into the buffer
- [x] Caching of pickups need to take the text into account otherwise...`Sent Progressive Jump to meep!`
- [x] should add/try a call to hud update for all kind of tank pickups
- [x] offworld model
- [x] offworld major item sound and name (???)
- [x] collecting offworld dna increments dna collected in scenario
- [ ] multiworld starter preset


src/open_samus_returns_rando/files/custom/message_gui.lua
ThanatosGit, line 1: Would be nice to get this on the main screen but no idea how to do.
If it stays on the second screen some sort of "long strings" handling is needed. Idc abou the color etc. atm

src/open_samus_returns_rando/files/custom/scenario.lua
```diff
@@ -142,4 +169,14 @@ end
...
+
+function Scenario.ShowAsyncPopup(text, time)
+    MessageGUI.UpdateMultiworld(true, text)
+    Scenario.hideSFID = Game.AddGUISF(time, "Scenario.HideAsyncPopup", "")
+end
+
+function Scenario.HideAsyncPopup()
+    MessageGUI.UpdateMultiworld(false, "")
+    Scenario.hideSFID = nil
 end
```
ThanatosGit, line 182: Should be replaced by a more robust approach

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
```diff
@@ -42,18 +42,52 @@ function RandomizerPowerup.IncreaseItemAmount(item_id, quantity, capacity)
...
+    if actorOrName.sName ~= nil then
+        name = actorOrName.sName
+    -- metroids
+    else
+        name = actorOrName
+    end
+    -- remote pickups from other worlds
+    if name == nil then
+        return
+    end
```
ThanatosGit, line 85: That shouldn't be possible anymore because it is catched before. Also it should lead to an error in L76 if this can be true.

src/open_samus_returns_rando/multiworld_integration.py
```diff
@@ -0,0 +1,56 @@
...
+          Path.open(input_exheader, "rb") as original,
+          Path.open(out_exheader, "wb") as result
+        ):
+        content = exheader_ips.read()
+        patch = ips.Patch.load(content)
+        patch.apply(original, result)
+
+    # copy bps patch (don't ask me why the patch does not work as IPS format)
+    output_exefs.mkdir(parents=True, exist_ok=True)
+    shutil.copyfile(files_path().joinpath("exefs_patches", "code.bps"), output_exefs.joinpath("code.bps"))
```
ThanatosGit, line 56: No idea if the patches works on NTSC and/or on Luma.
Code patch shouldn't be a problem for NTSC but no idea about exheader.
Have neither of them => Cannot test. If it doesn't work on Luma, I have no idea anyways on what to do.

On a side note: I was on Citra Canary 2704 and it had a bug in the services. Updated to Canary 2798 and it worked, so this will be the only version I will support.

ThanatosGit: Conclusion is/was: 
Exheader needs to be provided by the user. Same patch works on NTSC and PAL. As I expect many users to input the wrong file, there is a stupid check function for it.
code patch can also use the same patch but Citra doesn't like the IPS format. We had to provide different BPS patch for NTSC and PAL but they do the same.

Atm no further work done on why it doesn't boot on luma.

dyceron: I should try this on luma now that we have the best patches

dyceron (approved):

PR #319, opened 2024-04-13
Disable the intro spawn trigger for the newborn alpha if already dead

### LINKED ISSUE #318: Alpha newborn can respawn
Haven't tried to reproduce yet but it was something like:
- "backdoored" alpha newborn
- much later in the seed came back again via backdoor
- alpha spawned again after triggering for the first time the melee tutorial thing next to the missile block

dyceron:
Fixes #318

### DIFF
--- a/src/open_samus_returns_rando/files/levels/s000_surface.lua
+++ b/src/open_samus_returns_rando/files/levels/s000_surface.lua
@@ -76,7 +76,7 @@ function s000_surface.InitFromBlackboard()
     Game.SaturateSpawnGroup("SpawnGroup008")
   end
   if Scenario.ReadFromBlackboard("AlphaIntroPlayed", false) then
-    Game.DisableEntity("TG_Intro_Alpha")
+    Game.DisableTrigger("TG_Intro_Alpha")
   end
   -- if not Blackboard.GetProp("DEFEATED_ENEMIES", "Queen") or not (Blackboard.GetProp("DEFEATED_ENEMIES", "Queen") > 0) then
   Game.DisableEntity("LE_Queen_Door")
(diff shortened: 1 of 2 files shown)


dyceron:
Tested this a couple of times and the Alpha never respawned.

ThanatosGit:
Hmmm for me the alpha still respawns if you go there from the front. 
I added some output. As you can see the `Dead` callback is not called when you go there from the front and the trigger seems to be still active.
It's only called from the back and - as you can see from the output - after I killed the alpha the first time and being at the missile block.

@ThanatosGit Nice typos in the log messages 🙃

https://github.com/randovania/open-samus-returns-rando/assets/117127188/bf65c5b2-1dc8-4556-b7f7-e62cc091a6ad

image

dyceron:
That's so odd. This Alpha is always causing issues.

ThanatosGit:
https://github.com/randovania/open-samus-returns-rando/blob/f16dee6b920a23126e243a504b5c07fc907f01f8/src/open_samus_returns_rando/files/templates/metroid_template.lua#L37-L42

Oops, this should be `Game.DisableTrigger("TG_Intro_Alpha")`. Then no other change should be required.

dyceron:
If you could test this for me, that'd be great.

ThanatosGit (approved):
Works. The changes in `s000_surface.lua` are not required because `DisableTrigger` from the metroid template is persistent.

dyceron:
Wait so should anything else still be changed or those this good to go?

ThanatosGit:
> Wait so should anything else still be changed or those this good to go?

Isn't that twice the same case 😛?
Feel free to merge it that way.

The changs in `s000_surface.lua` didn't make a difference for me.
Kinda makes sense? 
We never set the `AlphaIntroPlayed` property because we completly block playing that cutscene.
The additional part in `s000_surface.OnEnter_Alpha_001_Dead` is alredy handled by the `metroid_template.lua`.
`OnDNAAbsorbAnimation` should also never be executed because we don't play this animation. That's also the reason why the `alpha_killed` part is in the `metroid_template.lua`.

PR #321, opened 2024-04-14
Support Elevator Shuffle

### LINKED ISSUE #23: Implement random elevators

dyceron:
Allows for the elevator destinations to be shuffled. Fixes #23 

TODO:

- [x] Add default elevator destinations so elevators not specified in a supplied JSON don't softlock the game when used
  - Either add to the schema or some other file

### DIFF
--- a/src/open_samus_returns_rando/files/custom/scenario.lua
+++ b/src/open_samus_returns_rando/files/custom/scenario.lua
@@ -3,6 +3,7 @@ Game.ImportLibrary("system/scripts/guilib.lua", false)
 Game.ImportLibrary("system/scripts/cosmetics.lua", false)
 
 Game.DoFile("system/scripts/room_names.lua")
+Game.DoFile("system/scripts/elevators.lua")
 
 Scenario = Scenario or {}
 setmetatable(Scenario, {__index = _G})
@@ -142,4 +143,14 @@ end
 
 function Scenario.UpdateRoomName(new_subarea)
   Game.AddSF(0.0, "RoomNameGui.Update", "s", new_subarea)
+end
+
+function Scenario.RandoOnElevatorUse(from_actor, _ARG_1_, _ARG_2_)
+  local destination = ElevatorDestinations[Scenario.CurrentScenarioID][from_actor.sName]
+  if Scenario.CurrentScenario == destination.scenario then
+    Game.AddGUISF(0.5, GUI.ElevatorStartUseActionStep2InnerArea, "")
+  else
+    Game.AddGUISF(0.5, GUI.ElevatorStartUseActionStep2InterArea, "")
+  end
+  Elevator.Use("c10_samus", destination.scenario, destination.actor, _ARG_2_)
 end
--- a/src/open_samus_returns_rando/files/custom/scenario.lua
+++ b/src/open_samus_returns_rando/files/custom/scenario.lua
@@ -147,10 +147,6 @@ end
 
 function Scenario.RandoOnElevatorUse(from_actor, _ARG_1_, _ARG_2_)
   local destination = ElevatorDestinations[Scenario.CurrentScenarioID][from_actor.sName]
-  if Scenario.CurrentScenario == destination.scenario then
-    Game.AddGUISF(0.5, GUI.ElevatorStartUseActionStep2InnerArea, "")
-  else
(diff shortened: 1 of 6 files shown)


src/open_samus_returns_rando/files/custom/scenario.lua
```diff
@@ -142,4 +143,14 @@ end
 
 function Scenario.UpdateRoomName(new_subarea)
   Game.AddSF(0.0, "RoomNameGui.Update", "s", new_subarea)
+end
+
+function Scenario.RandoOnElevatorUse(from_actor, _ARG_1_, _ARG_2_)
+  local destination = ElevatorDestinations[Scenario.CurrentScenarioID][from_actor.sName]
+  if Scenario.CurrentScenario == destination.scenario then
```
dyceron, line 150: Looking at this again, I'm not sure if this is ever valid, because I don't think `Scenario.CurrentScenario` exists. If it does exist, I don't think it's achieving whar we want. But I think it's just always ignored.

duncathan: if it doesn't exist then nothing in this function would work correctly

duncathan: oh nvm I misread. you meant to use `Scenario.CurrentScenarioID`

dyceron: My issue is what it's trying to achieve. All the subareas are different scenarios, so this would also be the else condition

dyceron: In my testing though, the elevators still seemed to use their vanilla inter vs inter callback, since that is referenced in the actors data itself, so maybe this isn't even necessary to have here.

src/open_samus_returns_rando/files/schema.json
```diff
@@ -114,21 +114,31 @@
...
-            "items": {
-                "type": "object",
-                "properties": {
-                    "teleporter": {
-                        "$ref": "#/$defs/actor_reference_with_layer"
-                    },
-                    "destination": {
-                        "$ref": "#/$defs/actor_reference"
+            "type": "object",
+            "description": "A dictionary of dictionaries mapping scenario and elevator actor to a connecting elevator",
```
ThanatosGit, line 118: Please add a `"default": {},` to it. 
Otherwise all patcher.json files break which is unecessary. (And it would also require to add an empty field from randovania etc....and to be totally nitpicky, it would technically be also version 2.0.0 then 😄)

src/open_samus_returns_rando/files/custom/scenario.lua
ThanatosGit: If this is required, yeah, you would need to use `Scenario.CurrentScenarioID`.
We have something similiar with the DNA mapping per area, which just uses the the first 3 characters. Of course this would require one stupid special case between s040 and s050.

ThanatosGit (changes requested):

src/open_samus_returns_rando/files/custom/scenario.lua
ThanatosGit: Oh wait, if you really need it you could actually use the exact same mapping.
You could use `Init.tScenarioMapping[Scenario.CurrentScenarioID]` and compare it with `Init.tScenarioMapping[destination.scenario]`

dyceron: I'm pretty confident it's not needed.

src/open_samus_returns_rando/files/schema.json
dyceron: Will do

src/open_samus_returns_rando/files/custom/scenario.lua
dyceron: I'll just test this again without these lines and see how it functions. If it behaves like it should in vanilla then I'll remove it. The actor might handle it already via the callback to the function in the elevator lua

dyceron: So after testing this more, it doesn't seem to make a difference if you use `Inner` or `Inter` here. Both are functioning the same, so something else might be at play which displays the minimap on the lower screen when an elevator is used. Not having any GUI line there does crash the game eventually, so I'll just use `Inter` as it's the safest option for now.

ThanatosGit (approved):

PR #323, opened 2024-04-20
Add custom title screen texture

dyceron:
Updates the title screen texture to mention `Randomizer`

image

I wish there was a better way to add this without having to just copy the whole file. I tried replicating how the custom init/scenario work, but the custom texture wasn't loaded. I'm not exactly sure what I'm missing from this to make it work otherwise. This will be especially important for Door Lock Rando when we add custom shields which would need to edit the respective lua file to add it to the map screen.


ThanatosGit:
Do you want me to look at the lua thing or just keep it as it is for now?

ThanatosGit:
Is the `MIZER` part one pixel lower than `RANDO`? It somehow looks kinda off to me.
But the `Z` seems to be on the correct position.

dyceron:
> Is the `MIZER` part one pixel lower than `RANDO`? It somehow looks kinda off to me. But the `Z` seems to be on the correct position.

Looks like it is one pixel lower on the top in places. I'll have to fix it.

dyceron:
> Do you want me to look at the lua thing or just keep it as it is for now?

Technically it doesn't have to be fixed right now, but it'll need to be done anyway later on. This is the last main thing I want to have included for stable.

dyceron:
I updated the texture to fix the misalignment of the lettering, and to also make the colors more consistent with the original lettering. Should be good to go now. I'm ok with merging this as is, but we will have to clean up the lua stuff down the line. 
image

ThanatosGit (approved):

PR #325, opened 2024-04-27
Refactor `heat_room_patches.py`

dyceron:
Future proofing for dumb ideas™


src/open_samus_returns_rando/specific_patches/heat_room_patches.py
```diff
@@ -14,226 +22,87 @@ def _get_new_logic_shape(editor: PatcherEditor, points: int) -> Container:
...
-    new_logic_shape["data"]["polys"][0]["boundings"] =  total_boundings
-    points = new_logic_shape["data"]["polys"][0]["points"]
+    points[0]["x"] = point[0][0]
+    points[0]["y"] = point[0][1]
+    points[1]["x"] = point[1][0]
+    points[1]["y"] = point[1][1]
+    points[2]["x"] = point[2][0]
+    points[2]["y"] = point[2][1]
+    points[3]["x"] = point[3][0]
+    points[3]["y"] = point[3][1]
```
ThanatosGit, line 94: `points` and `point` is confusing because both is a list of points.
At that point (pun intended 😛), I would just move it to `_get_new_logic_shape`, pass the list of points instead of the amount and create them directly with the correct coordinates.

dyceron: Move the entire boundings and points section? Don't I have to reference the class still?

ThanatosGit: Yes, either that or pass the list of points like I mentioned. Which made me realise that the type is wrong.

If you want to pass the whole reference like you mentioned. Something like that (untested):
```python
def _get_new_logic_shape(editor: PatcherEditor, new_actor: NewHeatActor) -> Container:
    scenario_2b = editor.get_scenario("s025_area2b")
    ls_copy = copy.deepcopy(scenario_2b.raw["objects_c"]["LS_Heat_001"])
    example_point = ls_copy["data"]["polys"][0]["points"][0]
    ls_copy["data"]["polys"][0]["points"].clear()
    for point in new_actor.points:
        new_point = copy.deepcopy(example_point)
        new_point["x"] = point[0]
        new_point["y"] = point[1]
        ls_copy["data"]["polys"][0]["points"].append(new_point)
    ls_copy["data"]["polys"][0]["num_points"] = len(new_actor.points)
    return ls_copy
```

src/open_samus_returns_rando/specific_patches/heat_room_patches.py
```diff
@@ -14,226 +22,87 @@ def _get_new_logic_shape(editor: PatcherEditor, points: int) -> Container:
...
-    name_of_scenario = "s040_area4"
 
-    scenario_4 = editor.get_scenario(name_of_scenario)
-
-    total_boundings = ListContainer([-850.0, 300.0, 4250.0, 1950.0])
+class NewHeatActor(typing.NamedTuple):
+    scenario: str
+    position: list[float]
+    total_boundings: list[float]
+    points: list[float]
```
ThanatosGit, line 30: The type is wrong. The values used in `new_heat_actors` for that field are at the moment of type `tuple[list[float],list[float],list[float],list[float]]`.
I think changing the type and values in `new_heat_actors` to `points: list[tuple[float, float]]` would be better.

With the other change, we then could add non recantgular shapes for whatever reason.

ThanatosGit (changes requested):

dyceron:
Tested the heat room in area_3c and it works.

src/open_samus_returns_rando/specific_patches/heat_room_patches.py
```diff
@@ -1,239 +1,101 @@
...
+    total_boundings: list[float]
+    points: list[tuple[float, float]]
+    entity_groups: list[str]
+    trigger_name: str = "TG_Heat_Rando_001"
+    logic_shape_name: str = "LS_Heat_Rando_001"
+
+new_heat_actors = [
+    NewHeatActor(
+        "s033_area3b", [1000.0, 5000.0, 0.0], [-500.0, 3800.0, 3900.0, 6700.0],
+        ([-500.0, 3800.0], [3900.0, 3800.0], [3900.0, 6700.0], [-500.0, 6700.0]),
```
ThanatosGit, line 27: Dynamically typed languages are fun 😀
The type in line 19 is correct but you forgot to change the values e.g. here in 27.
The `(` +`)` define a tuple and the `[` + `]` define a list. To match the type they should be used the exact opposite way.
To get `points: list[tuple[float, float]]` it should be `[(-500.0, 3800.0), (3900.0, 3800.0), (3900.0, 6700.0), (-500.0, 6700.0)],`

But it already works because you can access a tuple and a list the same way / with the same syntax.
Hence: Dynamically typed languages are fun

ThanatosGit (changes requested):

ThanatosGit (approved):

PR #326, opened 2024-04-27
Refactors the custom Metroid checkpoints

dyceron:
The custom Metroid checkpoints now use a class instead of being individually added. Honestly, this pr might be overkill since I doubt we'd actually add more checkpoints, but could be useful in the future for adding extra boss checkpoints. Either way, cleans up the code a bit so that's a plus.

### DIFF
--- a/src/open_samus_returns_rando/specific_patches/metroid_patches.py
+++ b/src/open_samus_returns_rando/specific_patches/metroid_patches.py
@@ -127,10 +127,8 @@ def add_startpoints(editor: PatcherEditor, new_startpoint: NewStartPoint):
         scenario_file.raw.actors[5][new_startpoint.st_name]["rotation"][1] = new_startpoint.st_rotation
         if new_startpoint.st_out_name is not None:
             scenario_file.raw.actors[5][new_startpoint.st_out_name]["rotation"][1] = (
-            -abs(new_startpoint.st_rotation)
-            if new_startpoint.st_rotation == 90
-            else abs(new_startpoint.st_rotation)
-        )
+                -90 if new_startpoint.st_rotation == 90 else 90
+            )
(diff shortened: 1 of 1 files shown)


src/open_samus_returns_rando/specific_patches/metroid_patches.py
```diff
@@ -53,259 +54,95 @@ def _patch_metroids(editor: PatcherEditor):
...
+
+    # Rotate the startpoint actors
+    if new_startpoint.st_name is not None:
+        scenario_file.raw.actors[5][new_startpoint.st_name]["rotation"][1] = new_startpoint.st_rotation
+        if new_startpoint.st_out_name is not None:
+            scenario_file.raw.actors[5][new_startpoint.st_out_name]["rotation"][1] = (
+            -abs(new_startpoint.st_rotation)
+            if new_startpoint.st_rotation == 90
+            else abs(new_startpoint.st_rotation)
+        )
```
ThanatosGit, line 133: if the value is 90 it does `-abs(new_startpoint.st_rotation)` but this is always just `-90`.
The whole ternary operator is just doing `-1 * new_startpoint.st_rotation`, isn't it?

dyceron: Yes it's always the opposite of the standard actor.

dyceron: I guess I could just hardcode the out actor's rotation based on the rotation value given. Would be simpler.

ThanatosGit (approved):

src/open_samus_returns_rando/specific_patches/metroid_patches.py
ThanatosGit: Don't know what you had against `-1 * new_startpoint.st_rotation` but whatever.

PR #329, opened 2024-04-28
Hotfix for Beam Block in Surface being deleted

dyceron:
Classic how in testing before this didn't come up, but I managed to pull it off from another CC instead 🙃


ThanatosGit (approved):

ThanatosGit:
Btw if you want this to be in for first rdv stable release, I would highly recommend to cherry pick solely this fix.

dyceron:
That would be nice. Not really sure how that works though.

PR #337, opened 2024-05-09
Fix creature doors not being removed

dyceron:
🙃


src/open_samus_returns_rando/specific_patches/door_patches.py
```diff
@@ -39,7 +39,7 @@
     "unk2": 1,
     "unk3": 0,
     "args": Container({
-        "601445949": Container({
```
ThanatosGit, line 42: Oopsie.
I was curious what is the string for this CRC32 and it's "CallbackName". Could use the string instead of the cryptic value.

ThanatosGit (approved):

PR #349, opened 2024-05-19
Fully ensure PBs can't be used when entering the Diggernaut fight

### LINKED ISSUE #348: Diggernaut PB Disable not covering all cases
If you start the fight, then reload to checkpoint and use a pb before reentering the fight, pbs are "disabled" but can still be used in the fight.

dyceron:
Fixes #348


ThanatosGit (approved):
lmao, what? why they keep being selected only after reset to cp but not before?

dyceron:
It turns out it has nothing to do with reloading. Entering the fight with less than max pbs would cause this to occur.

ThanatosGit:
> It turns out it has nothing to do with reloading. Entering the fight with less than max pbs would cause this to occur.

I played around with it earlier and I was able to also do it with max pbs. It's just morph and hold R while going through the door.

Issue #340, opened 2024-05-11
Add py.typed marker


ThanatosGit:
If done, the next patcher bump in Randovania should remove the `type: ignore` in the `game_exporter.py`

PR #354, opened 2024-05-21
Add py.typed marker

ThanatosGit:
For #340 (can be closed when the rdv side is done as described in the same issue)


dyceron (approved):
Not really sure if this is supposed to be empty or not, but I assume this will be filled later.

ThanatosGit:
Yeah, it is empty. It's just a marker for mypy. Need to look if this is actually packaged, so I merge and check what is pushed to test pypi (too lazy to do it local)

PR #355, opened 2024-05-21
Support for new MEDS

ThanatosGit:
Missing a (unreleased) MEDS bump


dyceron (approved):
Looks good to me. I'm pretty sure we don't change anything else, so this should be all of it.

PR #357, opened 2024-05-24
Fix MS bug

ThanatosGit:


dyceron (approved):
I wonder if it's worth combining all the area specific patches into the same function down the road

henriquegemignani:
> I wonder if it's worth combining all the area specific patches into the same function down the road

A dedicated function for each patch makes it clear what's being done. Look at how easy it is to know what `apply_static_fixes` does, for example.

dyceron:
That's fair

PR #358, opened 2024-05-27
Add support for adding custom door actors

dyceron:
New doors can now be easily added anywhere in the game. Currently used to add new doors to 3-tile tall open passages for Door Lock Rando. Add new doors via the schema.
image


src/open_samus_returns_rando/specific_patches/door_patches.py
```diff
@@ -466,6 +464,13 @@ def patch_minimap(
             # breaks minimap updating when shooting it from the left side but fixes itself after scan or reload
             left_tile_icon.actor_name = ''
 
+        # special case for the single tile save station in area 1
+        if scenario_name == "s010_area1" and actor_name in {"RandoDoor_004", "RandoDoor_005"}:
```
ThanatosGit, line 468: These names will be defined in RDV?

src/open_samus_returns_rando/specific_patches/door_patches.py
```diff
@@ -569,17 +617,18 @@ def add_custom_shields(editor: PatcherEditor, new_shield: NewShield) -> None:
...
+    # Add custom door actors
+    add_custom_doors(editor, custom_doors)
 
-    set_of_jsons = {json.dumps(d, sort_keys=True) for d in door_patches}
-    door_patches_set = [json.loads(t) for t in set_of_jsons]
+    # Create new door types
+    for new_shield in new_shields:
+        add_custom_shields(editor, new_shield)
 
-    for door in door_patches_set:
```
ThanatosGit, line 584: I prefer to keep it:
1. It's much faster to not patch every door twice
2. I'm pretty sure the way I wrote the minimap stuff will not work with the duplicates. Do you tested this?

ThanatosGit (changes requested):

src/open_samus_returns_rando/specific_patches/door_patches.py
dyceron: Yes they are

src/open_samus_returns_rando/specific_patches/door_patches.py
dyceron: My current wip branch on the rdv side which adds all the new doors makes changes to the door_patches function to not export duplicates. So the functionality is the same as before. The pr will be live soon so you'll seen it then.

ThanatosGit (approved):

PR #362, opened 2024-06-02
Add async popup windows

ThanatosGit:
currently look like this:

https://github.com/randovania/open-samus-returns-rando/assets/117127188/c34004e3-dafe-4d4f-87fb-58078702e857


ThanatosGit:
lol? Where did the energy and aeion came from at 0:46 seconds?

dyceron:
Ooh this looks good. Could you make it match how the vanilla popups look, and possibly change the color so it doesn't get confused with it?

ThanatosGit:
Hmmm no idea if can just change the color which means laying a color over a texture. At the very least, I could just change the color in the texture file but...
Actually, my intention was to make it later exactly look like vanilla. All the menu, ui etc, has this blueish color, so I prefer to keep it like that.
Also want to disable the vanilla tutorial stuff such that it won't overlay.

Only downside atm is the pickups which toggle back and forth the Idle screen because that also disables all UI in USSEF.

dyceron:
We could easily remove the tutorial popups.  I think they are all triggers (if not then via lua).

ThanatosGit:
Now looks like this. Should be good. Will remove the tutorial thing in a different PR.
image

dyceron:
Looks great. Is it possible to have the top line be vertically centered if the bottom line is empty, just like the tutorial popups?

ThanatosGit:
Ok, done.
image
image

dyceron (approved):

Issue #75, opened 2023-08-30
Progressive Models do not animate


dyceron:
https://github.com/randovania/open-samus-returns-rando/assets/38679103/d873d0cb-0c72-4463-8671-88e5a361a4a1

ThanatosGit:
Chances are pretty low that this will ever change IMO.
Dread also has no effects or animation on progressives and we are using the same component here to achieve it.

ThanatosGit:
As this is very likely an issue / bug which will stay open forever, we may just make a list of items we want to rotate? Iirc Wave Beam is facing towards the screen but all other stages of progressive beam are fine.
Can we rotate from lua? I don't want to implement a animation on every frame because of the heavy load but it would be nice to do it when we update the progressive model.
Otherwise the solution won't work anyways b

dyceron:
Closing due to merged pr and the very low likelihood of this ever changing.

PR #364, opened 2024-06-03
Rotate Wave Beam and High Jump Boots if progressive

dyceron:
If progressives are enabled and the model is either Wave Beam or High Jump Boots, the model will now face to the right to be more easily identifiable.

Related to #75 (Should close the issue after this)

Wave Beam
image

High Jump Boots
image


ThanatosGit (approved):
Technically, that also rotates varia suit but shouldn't be an issue because of how the varia suit model looks like

dyceron:
Oh shoot I forgot about Varia. I'll test it later and adjust this if needed.

PR #367, opened 2024-06-04
Update Reserves without Idle Screen

dyceron:
Works with both current and async item popups

Current

https://github.com/randovania/open-samus-returns-rando/assets/38679103/e9bfa89b-fe21-4c3e-a2b3-25257d29b1e4

Async

https://github.com/randovania/open-samus-returns-rando/assets/38679103/325b22b8-cc58-4c70-91bf-55b5130d9c75


ThanatosGit (approved):
nice

PR #372, opened 2024-06-10
Add mapping for pickups and map icons

### LINKED ISSUE #292: Create mapping for pickups and map icons
Currently, all pickups just use an open circle on the map. As a QoL, we could change this so items would use their respective map icons. Pickups and the map would have to be linked up together somehow. Items could show up as the open circle when the map hasn't been visited (Scan Pulse is used) and the actual item icon when visited. This is vanilla behavior and would be nice to replicate.

Since `itemsphere` always display regardless when Scan Pulse is used, we might need to make a new itemsphere icon (could just copy the current one or somehow figure out how to make it not show up when not visited) and call that. Also would be nice to add an icon for DNA.

dyceron:
Updates the minimap to display the each pickups' respective icon. Currently is always enabled. Should this be a toggle?

Things to mention:
- Nothing items and Powerups share the same itemsphere icon. Those two groups probably should get unique icons at some point. Icons for each major might be overkill and hard to make given the small size, so one icon is fine imo.
- Pickups that have custom blocks added will not update the minimap when broken. The rest of the hidden items will update when broken, like in vanilla.
- Custom icons are always visible even when in breakable blocks, revealing the pickup too early. I made them generic to deal with this.
- Queen pickup is always generic as the item would be revealed by default.

Fixes #292 

image

Adds new icons for DNA and Offworld models as well.

image


Miepee:
> Custom icons are always visible even when in breakable blocks, revealing the pickup too early. I made them generic to deal with this.

Is it possible to do that breaking the block changes the minimap icon to be non-generic? I feel like otherwise people might confuse it for a bug.

dyceron:
I don't know where that happens. If I don't change this, then most of the hidden items will change and a few won't and will be wrong.

ThanatosGit:
Can't remember how it works: Does it reveal the map icon by just using scan pulse?

dyceron:
Items are open circle until the tile is visited. I need fi recheck the custom icons to see how they behave.

dyceron:
> Items are open circle until the tile is visited. I need to recheck the custom icons to see how they behave.

They were broken, but now they are only revealed when the tile is visited.

tests/test_files/starter_preset_patcher.json
```diff
@@ -272,12 +273,12 @@
         },
         {
             "pickup_type": "actor",
-            "caption": "Morph Ball acquired.",
+            "caption": "Sent Speed Booster to meep.",
```
ThanatosGit, line 276: Should have been a progressive jump 😥
Was it intentional to change the **starter_preset** file with different start point, multiworld stuff and the debug thing `reveal_map_on_start` ?

dyceron: Oh whoops no that was supposed to be reverted

ThanatosGit (approved):

PR #381, opened 2024-06-20
Pickup Tile Updates

dyceron:
- Refactor static changes for items and doors
- Explicitly check for most models to update the map accordingly


ThanatosGit (approved):
Unrelated to this PR, but whenever I see a change in the door_patcher file, I think that it should be heavily refactored.

PR #385, opened 2024-06-23
Add more offworld models

dyceron:
From left to right:
- Power Grip (AM2R)
- Speed Booster (AM2R, Dread)
- Seeker Launcher (Echoes)/Storm Missile (Dread)/Wavebuster (Prime)
- Flamethrower (Prime)
- Ice Spreader (Prime)/Sunburst (Echoes)
- Darkburst (Echoes)
- Sonic Boom (Echoes)

https://github.com/randovania/open-samus-returns-rando/assets/38679103/f089889b-228b-4b37-8a1e-1354b96e80ab


dyceron:
Feel like Darkburst should be even darker, but unsure

Miepee:
I feel like "here are recolors of a missile launcher" may not be the best accessibililty wise

dyceron:
Well we don't exactly have the ability to make custom models

Miepee:
Which makes me wonder if it would be better to either keep using generic offworld model for them, or to have a generic missile upgrade.

dyceron:
I'd rather it not just be generic offworld, so I guess the latter option. Recolored missile still?

dyceron:
Closing in favor of #391

PR #387, opened 2024-06-25
Add hint for Baby location on collected all DNA textbox

dyceron:
Reads a string from the configuration which replaces the second sentence with a Baby hint. Will be configurable in Randovania to choose between no hint (current behavior), area only, or precise. Useful as a last resort if in go mode otherwise, or if a Baby hint is not provided

image

### DIFF
--- a/src/open_samus_returns_rando/files/schema.json
+++ b/src/open_samus_returns_rando/files/schema.json
@@ -386,19 +386,34 @@
         "hints": {
             "type": "array",
             "items": {
-                "type": "object",
-                "properties": {
-                    "accesspoint_actor": {
-                        "$ref": "#/$defs/actor_reference"
-                    },
-                    "text": {
-                        "type": "string"
+                "if": {
+                    "properties": {
+                        "baby_metroid_hint": {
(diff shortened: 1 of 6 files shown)


dyceron:
I'm pretty sure the schema is wrong.

ThanatosGit (commented):
Is there a reason that both is in the `hints` array?
The usage in `src/open_samus_returns_rando/specific_patches/hint_patches.py` is now "Process everything besides baby hint" and in `src/open_samus_returns_rando/lua_editor.py` it's "Process nothing besides baby hint".
It looks like it shouldn't be in the same object at all.

dyceron:
It's in `hints` for simplicity. I was just basing this on how the export would be in the PDF in RDV to avoid duplicating things. Though I guess I'm already doing that by adding the Baby hint so that's irrelevant.

The other changes follow suit, but this is making this harder. I guess it'd be much easier and simpler to have another field in the config instead of grouping it.

dyceron:
Moved the hint to its own field for simplicity

ThanatosGit (approved):

PR #388, opened 2024-06-25
Add Speed Booster and Powergrip offworld models

dyceron:
- Power Grip (AM2R)
- Speed Booster (AM2R, Dread)

image


src/open_samus_returns_rando/files/schema.json
```diff
@@ -665,7 +665,6 @@
                 "ITEM_RESERVE_TANK_LIFE",
                 "ITEM_RESERVE_TANK_MISSILE",
                 "ITEM_RESERVE_TANK_SPECIAL_ENERGY",
-                "ITEM_OFFWORLD",
```
ThanatosGit, line 668: I'm in a rush: Why is this removed? We still have a generic offworld model which is the RDV logo icon thing, don't we?

tests/test_files/item_models_test.json
```diff
@@ -697,8 +697,8 @@
             "resources": [
                 [
                     {
-                        "item_id": "ITEM_OFFWORLD",
-                        "quantity": 1
```
ThanatosGit, line 701: Same question here (even though this is just a test file for us)

src/open_samus_returns_rando/files/schema.json
dyceron: Because I remembered that offworld models use `ITEM_NONE`, so this item_id is technically never used. Obviously we could use it in the tests, but None with 0 quantity works just fine.

dyceron: Generic offworld is unaffected

ThanatosGit: Oh, OK. So, it is `ITEM_NONE` just with a different model. Totally makes sense

ThanatosGit (approved):

src/open_samus_returns_rando/files/schema.json
dyceron: Yep

PR #397, opened 2024-07-05
Enable A6 Omega by default

dyceron:
Relevant if starting in upper A6

### DIFF
--- a/src/open_samus_returns_rando/files/levels/s070_area7.lua
+++ b/src/open_samus_returns_rando/files/levels/s070_area7.lua
@@ -91,6 +91,8 @@ s070_area7.tDNAScanLandmarks = {
   }
 }
 function s070_area7.InitFromBlackboard()
+  Game.SetSubAreaCurrentSetup("collision_camera_037", "Omega_Enabled", true)
+  Scenario.WriteToBlackboard("OmegaDiscovered", "b", true)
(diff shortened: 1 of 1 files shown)


src/open_samus_returns_rando/files/levels/s070_area7.lua
```diff
@@ -140,8 +142,6 @@ function s070_area7.OnHazarousPoolDrain(_ARG_0_)
   --  Game.SetSubAreaCurrentSetup("collision_camera_044", "Hazardous_Phase_03", true)
   --  Game.SetSubAreaCurrentSetup("collision_camera_042", "Hazardous_Phase_03", true)
   --end
-  Game.SetSubAreaCurrentSetup("collision_camera_037", "Omega_Enabled", true)
-  Scenario.WriteToBlackboard("OmegaDiscovered", "b", true)
```
ThanatosGit, line 144: I really thought `OnHazarousPoolDrain` is called by default but looks like it doesn't and now I wonder why the whole thing is working at all.
All the other setups are in the lower section? So it is called at one point when going to lower A6?

ThanatosGit (approved):

src/open_samus_returns_rando/files/levels/s070_area7.lua
dyceron: I'm not sure about the lowest section, but tou had to enter the Hazardous area to reach the Omega before, so it just worked.

PR #398, opened 2024-07-05
Support Music Shuffle

### LINKED ISSUE #69: Add Music Shuffle
Ideally have two options: 
- Shuffle all music together 
- Shuffle areas and metroid/bosses separately

dyceron:
Takes a dictionary of tracks as input and replaces the data with a new specified track

Fixes #69

### DIFF
--- a/src/open_samus_returns_rando/files/schema.json
+++ b/src/open_samus_returns_rando/files/schema.json
@@ -499,7 +499,7 @@
                     }
                 },
                 "music_shuffle_dict": {
-                    "description": "A dictionary mapping music tracks to the output of music mode.",
+                    "description": "A dictionary mapping of music tracks where the keys get swapped with the values.",
                     "type": "object",
                     "patternProperties": {
                         ".*": {
--- a/src/open_samus_returns_rando/files/schema.json
+++ b/src/open_samus_returns_rando/files/schema.json
@@ -497,6 +497,17 @@
                             }
                         }
                     }
+                },
+                "music_shuffle_dict": {
+                    "description": "A dictionary mapping music tracks to the output of music mode.",
+                    "type": "object",
+                    "patternProperties": {
+                        ".*": {
+                            "type": "object",
+                            "$ref": "#/$defs/music_list"
+                        }
+                    },
+                "default": {}
                 }
             },
             "default": {}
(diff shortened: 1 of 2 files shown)


src/open_samus_returns_rando/specific_patches/cosmetic_patches.py
```diff
@@ -33,3 +34,10 @@ def lua_cosmetics(configuration: dict) -> str:
     }
 
     return lua_util.replace_lua_template("cosmetics.lua", replacement)
+
+
+def music_shuffle(editor: PatcherEditor, configuration: dict) -> None:
+    for original, new in configuration["music_shuffle_dict"].items():
+        original_track = f"sounds/streams/music/{original}.bcwav"
+        new_track = editor.get_raw_asset(f"sounds/streams/music/{new}.bcwav")
+        editor.replace_asset(original_track, new_track)
```
Miepee, line 43: This currently has some unexpected behaviour with something like
```json
"k_crateria99": "boss_go",
"boss_go": "k_crateria99"
```

src/open_samus_returns_rando/specific_patches/cosmetic_patches.py
```diff
@@ -33,3 +34,14 @@ def lua_cosmetics(configuration: dict) -> str:
     }
 
     return lua_util.replace_lua_template("cosmetics.lua", replacement)
+
+
+def music_shuffle(editor: PatcherEditor, configuration: dict) -> None:
```
Miepee, line 39: can you early return? something like
```py
if configuration["music_shuffle_dict"] == {}:
  return
```

src/open_samus_returns_rando/files/schema.json
```diff
@@ -497,6 +497,17 @@
                             }
                         }
                     }
+                },
+                "music_shuffle_dict": {
+                    "description": "A dictionary mapping music tracks to the output of music mode.",
```
ThanatosGit, line 502: I don't get what this has to do with `music mode`. 
It swaps one track with the other.

dyceron: I don't understand the confusion.

Miepee: Music mode has no effect here and is completely irrelevant. A more accurate description would be something along the lines of "A dictionary mapping, where the key is a track that gets swapped with the value"

dyceron: Fwiw this was made before the rdv pr, so this is just a small wording error on my part. I'll fix it.

ThanatosGit (approved):

PR #399, opened 2024-07-08
A7 Save Station Blocks Edge Case

dyceron:
There was another model unaccounted for 🙃. Is there a better way to simplify this whole function? It's rather messy.

### DIFF
--- a/src/open_samus_returns_rando/specific_patches/static_fixes.py
+++ b/src/open_samus_returns_rando/specific_patches/static_fixes.py
@@ -320,7 +320,7 @@ def disable_vignettes(editor: PatcherEditor) -> None:
         ],
         # Middle Save Station Water Bomb Blocks
         "s090_area9": [
-            {"block_group": 66, "cc": ["007"], "vignette_models": [3224282959]}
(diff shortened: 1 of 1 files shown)


src/open_samus_returns_rando/specific_patches/static_fixes.py
```diff
@@ -320,7 +320,7 @@ def disable_vignettes(editor: PatcherEditor) -> None:
         ],
         # Middle Save Station Water Bomb Blocks
         "s090_area9": [
-            {"block_group": 66, "cc": ["007"], "vignette_models": [3224282959]}
+            {"block_group": 66, "cc": ["007"], "vignette_models": [3224282959, 1]}
```
ThanatosGit, line 323: Somehow I thought that this is the result of a CRC. Is this really just `1`?

ThanatosGit (approved):

src/open_samus_returns_rando/specific_patches/static_fixes.py
dyceron: Yeah it's part of a separate group in that sg

PR #400, opened 2024-07-08
Add option to prevent breakable blocks from respawning

### LINKED ISSUE #295: Adjust respawn timers for blocks in tunnels
All shot blocks in tunnels could just be changed to never respawn, which would help remove softlocks without bombs or pbs. Bomb blocks could also be changed but these might need to looked at in a case by case basis depending on what might be affected logically.

dyceron:
Excludes `weight` blocks

Fixes #295

### DIFF
--- a/src/open_samus_returns_rando/files/schema.json
+++ b/src/open_samus_returns_rando/files/schema.json
@@ -297,6 +297,11 @@
                     "type": "boolean",
                     "description": "Removes the wall between the Queen and Baby, allowing access into Area 8 from Surface West.",
                     "default": false
+                },
+                "respawning_blocks": {
+                    "type": "boolean",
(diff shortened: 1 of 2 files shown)


ThanatosGit:
Isn't that really bad for rooms like that? 
image

dyceron:
Ah yeah it is. Bleh this option isn't going to be fun to have. Would need exclusions but that sounds a like a lot of work.

dyceron:
Maybe this should be restricted to bomb blocks only?

ThanatosGit:
It would require me to remember if there are similiar situations with bomb blocks. 🫠

It also came to my mind that this one here would be completly trivialized with the current state. Not a big fan of this option and writing down all the exclusion is really meh
image

dyceron:
Yeah this option isn't great imo.

dyceron:
Closing until a better solution is considered

PR #403, opened 2024-07-09
Re-enable SpecialEnergyCloud for powerups

### LINKED ISSUE #104: Restore SpecialEnergyOrb to Major Pickup locations

dyceron:
Re-enables the `SpecialEnergyCloud` for each powerup in a Chozo or Aeion Statue. Special handling was needed for Super Missiles and the Aeion Abilities due to naming issues.

Fixes #104

### DIFF
--- a/src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
+++ b/src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
@@ -97,17 +97,31 @@ function RandomizerPowerup.MarkLocationCollected(actorOrName)
     Blackboard.SetProp(playerSection, propName, "b", true)
 end
 
+local AeionAbilityClouds = {
+    {"s000_surface", "LE_SpecialAbility_ScanningPulse"},
+    {"s028_area2c", "LE_SpecialAbility_EnergyShield"},
+    {"s030_area3", "LE_SpecialAbility_EnergyWave"},
+    {"s060_area6", "LE_SpecialAbility_PhaseDisplacement"},
+}
+
 function RandomizerPowerup.ActivateSpecialEnergy(actorOrName)
     local name = actorOrName.sName
-    local cloud = ""
-    if name ~= nil then
-        if string.sub(name, 0, 8) == "LE_Power" then
-            cloud = string.sub(name, 11)
-        -- SuperMissile actor name has a typo, so explicity set the value of cloud
-        elseif name == "LE_PoweUp_SuperMissile" then
-            cloud = "_SuperMissile"
+    local cloud = "TG_SpecialEnergyCloud"
+    local trigger = cloud .. string.sub(name, 11)
+
+    -- Powerups
+    if string.sub(name, 0, 8) == "LE_Power" and trigger ~= nil then
+        SpecialEnergyCloud.ActivateSpecialEnergy(trigger)
+    -- The Super Missile actor name has a typo, so explicity activate its cloud
+    elseif name == "LE_PoweUp_SuperMissile" then
+        SpecialEnergyCloud.ActivateSpecialEnergy(cloud .. "_SuperMissile")
+    end
+
+    -- Aeion abilities
+    for _, mapping in ipairs(AeionAbilityClouds) do
+        if mapping[1] == Scenario.CurrentScenarioID and mapping[2] == name then
+            SpecialEnergyCloud.ActivateSpecialEnergy(cloud)
         end
-        SpecialEnergyCloud.ActivateSpecialEnergy("TG_SpecialEnergyCloud" .. cloud)
     end
 end
 
--- a/src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
+++ b/src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
@@ -51,7 +51,7 @@ function RandomizerPowerup.OnPickedUp(resources, actorOrName)
 
     Scenario.UpdateProgressiveItemModels()
     if actorOrName ~= nil then
-        RandomizerPowerup.GetActorName(actorOrName)
+        RandomizerPowerup.CheckActorOrName(actorOrName)
     end
     RandomizerPowerup.IncrementInventoryIndex()
     RL.UpdateRDVClient(false)
@@ -71,7 +71,7 @@ function RandomizerPowerup.PropertyForLocation(actorOrName)
     return "c_" .. actorOrName
 end
 
-function RandomizerPowerup.GetActorName(actorOrName)
+function RandomizerPowerup.CheckActorOrName(actorOrName)
     local name
     -- normal pickups
     if actorOrName.sName ~= nil then
@@ -111,10 +111,8 @@ function RandomizerPowerup.ActivateSpecialEnergy(name)
         if Game.GetEntity(trigger) ~= nil then
             SpecialEnergyCloud.ActivateSpecialEnergy(trigger)
         end
-    end
-
-    -- Aeion abilities
-    if string.sub(name, 0, 17) == "LE_SpecialAbility" and Game.GetEntity(name) ~= nil then
+    else
+        -- Aeion abilities
         SpecialEnergyCloud.ActivateSpecialEnergy(cloud)
     end
 end
--- a/src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
+++ b/src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
@@ -97,29 +97,22 @@ function RandomizerPowerup.MarkLocationCollected(actorOrName)
     Blackboard.SetProp(playerSection, propName, "b", true)
 end
 
-local AeionAbilityClouds = {
-    {"s000_surface", "LE_SpecialAbility_ScanningPulse"},
-    {"s028_area2c", "LE_SpecialAbility_EnergyShield"},
-    {"s030_area3", "LE_SpecialAbility_EnergyWave"},
-    {"s060_area6", "LE_SpecialAbility_PhaseDisplacement"},
-}
-
 function RandomizerPowerup.ActivateSpecialEnergy(actorOrName)
     local name = actorOrName.sName
     local cloud = "TG_SpecialEnergyCloud"
-    local trigger = cloud .. string.sub(name, 11)
 
     -- Powerups
-    if string.sub(name, 0, 8) == "LE_Power" and trigger ~= nil then
-        SpecialEnergyCloud.ActivateSpecialEnergy(trigger)
-    -- The Super Missile actor name has a typo, so explicity activate its cloud
-    elseif name == "LE_PoweUp_SuperMissile" then
-        SpecialEnergyCloud.ActivateSpecialEnergy(cloud .. "_SuperMissile")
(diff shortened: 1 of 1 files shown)


dyceron:
On a second look, since Supers need special handling like the arion, maybe I should just remove explicit check and add it to the table with the aeion.

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
```diff
@@ -96,6 +97,34 @@ function RandomizerPowerup.MarkLocationCollected(actorOrName)
     Blackboard.SetProp(playerSection, propName, "b", true)
 end
 
+local AeionAbilityClouds = {
```
ThanatosGit, line 100: `local` on file level doesn't make much sense, does it? I would just move it into the function because there is no reason to make it global.

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
```diff
@@ -96,6 +97,34 @@ function RandomizerPowerup.MarkLocationCollected(actorOrName)
...
+    {"s000_surface", "LE_SpecialAbility_ScanningPulse"},
+    {"s028_area2c", "LE_SpecialAbility_EnergyShield"},
+    {"s030_area3", "LE_SpecialAbility_EnergyWave"},
+    {"s060_area6", "LE_SpecialAbility_PhaseDisplacement"},
```
ThanatosGit, line 104: To me it looks like the names are unique, so there is no reason to have the area name + the check for it IMO.

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
```diff
@@ -96,6 +97,34 @@ function RandomizerPowerup.MarkLocationCollected(actorOrName)
...
+}
+
+function RandomizerPowerup.ActivateSpecialEnergy(actorOrName)
+    local name = actorOrName.sName
+    local cloud = "TG_SpecialEnergyCloud"
+    local trigger = cloud .. string.sub(name, 11)
```
ThanatosGit, line 110: This will throw a lua exception for metroids. `actorOrName` is either an actor or really just a name a.k.a. a string. In the later case `sName` doesn't exist on it, so `local name` is `nil` and you cannot use `nil` with `string.sub`.

For reference see the check in `MarkLocationCollected`. Probably worth refacotring the check out and do it before calling both methods such that both only get a string.

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
```diff
@@ -96,6 +97,34 @@ function RandomizerPowerup.MarkLocationCollected(actorOrName)
...
+
+    -- Powerups
+    if string.sub(name, 0, 8) == "LE_Power" and trigger ~= nil then
```
ThanatosGit, line 113: Trigger can never be `nil`

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
```diff
@@ -96,6 +97,34 @@ function RandomizerPowerup.MarkLocationCollected(actorOrName)
...
+        SpecialEnergyCloud.ActivateSpecialEnergy(trigger)
+    -- The Super Missile actor name has a typo, so explicity activate its cloud
+    elseif name == "LE_PoweUp_SuperMissile" then
+        SpecialEnergyCloud.ActivateSpecialEnergy(cloud .. "_SuperMissile")
+    end
```
ThanatosGit, line 118: If I understand correctly, the special case is just for the typo in `PowerUp`.
Fwiw, you could just find the last `_` and take the sub string from there.
Something like
```lua
if string.sub(name, 0, 7) == "LE_Powe" then
    local indexstart, indexend = string.find(name, "_", 4)
    local trigger = cloud .. string.sub(name, indexend)
    SpecialEnergyCloud.ActivateSpecialEnergy(trigger)
end
```

ThanatosGit (changes requested):

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
dyceron: I'm doing it this way because all of the aeion ability clouds are named the same, so getting one of these would activate the rest.

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
dyceron: It's definitely can be. Not every powerup has a cloud, so this filters those out before the call happens.

ThanatosGit: But the lines directly above doing this:
```lua
    local cloud = "TG_SpecialEnergyCloud"
    local trigger = cloud .. string.sub(name, 11)
```
`trigger` can never be `nil` because it will always contain at least "TG_SpecialEnergyCloud"

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
ThanatosGit: Hmmm but it doesn't activate it in a specific scenario, it's always just
```lua
SpecialEnergyCloud.ActivateSpecialEnergy(cloud)
```
where `cloud = "TG_SpecialEnergyCloud"`.
This only activates it in the scenario you are currently at.

dyceron: That's what the loop is for. If the scenarios don't match then it doesn't activate in the scenario

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
dyceron: Oh then it should probably be `.... and Game.GetEntity(trigger) ~= nil)`

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
ThanatosGit: Yes, but my point is that you will never find e.g. `"LE_SpecialAbility_ScanningPulse"` in a different scenario than `"s000_surface"`, so checking the scenario is not relevant.

dyceron: I guess I'll test without the check but i don't think it's gonna work

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
dyceron: Couldn't I say `name = actorOraname.sName if name ~= nil`?

dyceron: Ah I see the ch3ck in that function. I'll pull that out

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
ThanatosGit: I try it with other words:
The loop is currently:
```lua
    for _, mapping in ipairs(AeionAbilityClouds) do
        if mapping[1] == Scenario.CurrentScenarioID and mapping[2] == name then
            SpecialEnergyCloud.ActivateSpecialEnergy(cloud)
        end
    end
```
`name` is the name of the actor you just picked up and `mapping[2]` is one of the values `"LE_SpecialAbility_ScanningPulse", "LE_SpecialAbility_EnergyShield", "LE_SpecialAbility_EnergyWave", "LE_SpecialAbility_PhaseDisplacement"`

So, if you just picked up `LE_SpecialAbility_PhaseDisplacement` there is no way that the check  ` mapping[2] == name` is `true` without being in `s060_area6` because that actor name doesn't exist in any other scenario.
For all other pickups it's `false` because the name is gonna be something like `LE_Item_001` etc.

dyceron: Couldn't I just leave the checks but modify the name check to be get entity?

ThanatosGit: For what purpose?

dyceron: This is me just talking what might happen but I just need to test and see.

dyceron:
Changed what was needed and refactored `actororname`. Tested if the energy clouds activate properly and they do. I don't think I've missed anything.

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
```diff
@@ -51,7 +51,7 @@ function RandomizerPowerup.OnPickedUp(resources, actorOrName)
 
     Scenario.UpdateProgressiveItemModels()
     if actorOrName ~= nil then
-        RandomizerPowerup.MarkLocationCollected(actorOrName)
+        RandomizerPowerup.GetActorName(actorOrName)
```
ThanatosGit, line 54: Function name is super misleading. `Get` but it doesn't return an actor and also not a name. In fact, it doesn't return anything.

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
```diff
@@ -96,6 +101,24 @@ function RandomizerPowerup.MarkLocationCollected(actorOrName)
...
+    if string.sub(name, 0, 7) == "LE_Powe" then
+        local powerup = string.find(name, "_", 4)
+        local trigger = cloud .. string.sub(name, powerup)
+        if Game.GetEntity(trigger) ~= nil then
+            SpecialEnergyCloud.ActivateSpecialEnergy(trigger)
+        end
+    end
+
+    -- Aeion abilities
+    if string.sub(name, 0, 17) == "LE_SpecialAbility" and Game.GetEntity(name) ~= nil then
```
ThanatosGit, line 117: The entity check doesn't make sense here.

It will be only checked if the first part is `true` (which is) `string.sub(name, 0, 17) == "LE_SpecialAbility"`. In the second part of the `and` you can already be sure that the name starts with that and we wouldn't have been able to pick it up if the entity for the name doesn't exist.
You could check if the `cloud` exists but afaik `cloud` exists for every pickup starting with `LE_SpecialAbility"`.

ThanatosGit (changes requested):

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
dyceron: What do you suggest then?

ThanatosGit: I'm bad with names :D 
The main thing it does is actually calling the mark location and special energy function. Usually you name something after the stuff it does and not after what parameters it takes.

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
ThanatosGit: I said the "entity check" not the whole check :p.
Now, you are activating it even when you just pick up any minor item location.
The name check needs to stay.

dyceron: brain moment

PR #406, opened 2024-07-10
Disable low poly Samus if reloading from checkpoint during boss fight

### LINKED ISSUE #317: Reloading from Checkpoint while in Diggernaut fight doesnt reset low poly model
Maybe dying doesnt do it too? wasnt tested.

dyceron:
`OnPlayerDead` already handles disabling the low poly on death, but reloading from checkpoint while in a boss fight is not handled in vanilla. This ensures low poly is disabled when a scenario load happens.

Fixes #317

### DIFF
--- a/src/open_samus_returns_rando/files/levels/s070_area7.lua
+++ b/src/open_samus_returns_rando/files/levels/s070_area7.lua
@@ -108,6 +108,7 @@ end
 function s070_area7.OnReloaded()
 end
 function s070_area7.OnExit()
+  s070_area7.SetLowModelsVisibility(false)
 end
 function s070_area7.OnEnter_ActivationDNA_001()
   -- Game.OnDNAMechApproached("LE_ChozoUnlockAreaDNA_001", 1)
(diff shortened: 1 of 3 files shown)


ThanatosGit (commented):
What happens if you reload to save but the save is not in one of the 3 areas?

dyceron:
Ah that's a good point. I wonder if it's possible to generalize this and add it to the scenario.lua` instead?

ThanatosGit:
Sounds fine if reloading to save is really an issue. I assume it because the low poly stuff seems to be a "play session wide global".
Thinking about it, now I'm curious what happens currently if you go with gravity to Diggernaut and then reload to a save without gravity. Invisible Samus until you pick up gravity again?

dyceron:
Just tested swapping from low poly grav to power and power was normal, the switching back to gravity was normal poly.

dyceron:
Figured out how to disable when leaving scenarios in the last commit

ThanatosGit (approved):
Oh, that's nice. Didn't now that this exists.

PR #407, opened 2024-07-10
Surface West Music Changes

dyceron:
Update the music in Surface West to be Surface East 1 if DNA is not 39 and Baby is not collected, and then uses the regular theme once the conditions to access Ridley are met. The reason for doing this is to add yet another indicator if the seed can be finished.

Debated on doing the 2nd Surface theme to indicate its a later area, but I think the first theme fits better since Landing Site uses that theme.


ThanatosGit (approved):
Without looking it up, I have no idea what all the parameters are.
As long as they do "stop the currently played stream and loop the new one" it lgtm.

dyceron:
I pretty much just copied the music function call from a7 for the omegas and changed the the track and conditional.

PR #408, opened 2024-07-10
Rotate Arachnus pickup to face the correct direction

### LINKED ISSUE #10: Item from Arachnus faces the wrong way
If the item that Arachnus drops is a tank, it will face to the right. This does not affect Major Items because they rotate and are viewable from all angles.

image

dyceron:
Thanatos' original branch and changes. I just moved the file and replace the content like the other copied files.

Fixes #10


ThanatosGit (approved):
Just for the record: This solution is ass :)

dyceron:
😁

PR #410, opened 2024-07-12
Add time to patcher cli

dyceron:
Helpful information

### DIFF
--- a/src/open_samus_returns_rando/cli.py
+++ b/src/open_samus_returns_rando/cli.py
@@ -2,6 +2,7 @@ import argparse
 import json
 import logging
 import logging.config
+import time
 from pathlib import Path
 
 from open_samus_returns_rando import samus_returns_patcher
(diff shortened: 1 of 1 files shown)


src/open_samus_returns_rando/cli.py
```diff
@@ -60,9 +61,12 @@ def main() -> None:
...
 
+    start = time.time()
     samus_returns_patcher.patch_extracted(
         args.input_path,
         args.input_exheader,
         args.output_path,
         configuration,
     )
+    end = time.time()
+    print(f"Patcher took {end - start:.03f} seconds")
```
Miepee, line 72: Methinks this should not be in release builds

dyceron: It exists for odr so 🤷‍♂️

henriquegemignani: Randovania doesn't use this function and it's convenient when running manually.

Miepee: It still shows up in randovania builds via the cmd console that pops up on windows tho, no?

ThanatosGit (approved):

Issue #409, opened 2024-07-12
Add Chaos Options


dyceron:
Some ideas include, but are not limited to:
- Randomly submerged rooms (water, lava, acid??)
- Randomly heated rooms
- Block Types shuffle 
- Enemy shuffle (type, speed, size etc)

Miepee:
I think thanatos wanted to experiment at one point with an alpha-squad.
No idea how that could be doable, or even made nicely, but I just found the concept so funny i wanted to  throw it out here.

PR #414, opened 2024-07-14
Support shuffling block types

dyceron:
Changes all the blocks of a certain type to another type. Note that will be a chaos option in Randovania, so it will not be logical, but the groundwork for it is there if that ever happens.

Relevant to #409

### DIFF
--- a/src/open_samus_returns_rando/misc_patches/block_patches.py
+++ b/src/open_samus_returns_rando/misc_patches/block_patches.py
@@ -12,10 +12,8 @@ def patch_block_types(editor: PatcherEditor, configuration: dict) -> None:
         blocks = bmsbk.raw["block_groups"]
         for block in blocks:
             block_type = block["types"][0]
-            temp_type: str = ""
             for original_type, new_type in configuration["block_patches"].items():
                 # Check for the block type and change it to the new type
-                if original_type == block_type["block_type"] and temp_type == "":
-                    # Set the temp_type to the original_type to ensure the block doesnt get changed again
-                    temp_type = original_type
+                if original_type == block_type["block_type"]:
                     block_type["block_type"] = new_type
+                    break
--- a/src/open_samus_returns_rando/misc_patches/block_patches.py
+++ b/src/open_samus_returns_rando/misc_patches/block_patches.py
@@ -1,0 +1,21 @@
+from mercury_engine_data_structures.formats import Bmsbk
+from open_samus_returns_rando.constants import ALL_SCENARIOS
+from open_samus_returns_rando.patcher_editor import PatcherEditor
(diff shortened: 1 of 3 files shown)


src/open_samus_returns_rando/misc_patches/block_patches.py
```diff
@@ -0,0 +1,21 @@
...
+        bmsbk = editor.get_file(f"maps/levels/c10_samus/{scenario_name}/{scenario_name}.bmsbk", Bmsbk)
+        blocks = bmsbk.raw["block_groups"]
+        for block in blocks:
+            block_type = block["types"][0]
+            temp_type: str = ""
+            for original_type, new_type in configuration["block_patches"].items():
+                # Check for the block type and change it to the new type
+                if original_type == block_type["block_type"] and temp_type == "":
+                    # Set the temp_type to the original_type to ensure the block doesnt get changed again
+                    temp_type = original_type
```
ThanatosGit, line 20: Please remove the `temp_type` shenanigans and just use `break` to stop the inner `for` loop.

ThanatosGit (changes requested):
More general question: How is that supposed to work with the static changes?
Should this be pre- or post the static_fixes?
Does this even make a difference? The A1 teleporter crumbles doesn't really care what block type they are.

Can you destroy a crumble block in a one tile gap within a wall or will all those pickups be unacessible then?

Miepee:
> Can you destroy a crumble block in a one tile gap within a wall or will all those pickups be unacessible then?

A ~~weight~~ crumble block gets destroyed if you touch it, doesn't necessarily have to be from above. You can just jump into them, spider boost into them, probably morph into them doesn't really matter.

As example, dcr had a test seed where the power bomb blocks here were replaced with crumbles, and they just got destroyed by jumping into them from side
grafik

ThanatosGit:
They definitely don't get destroyed if touching them from below 😝.
Iirc there is even a small spot you could spider ball around the corner without them getting destroyed, that's why I asked explicitly about one tile gaps. But I assume it will be just fine and that you can destroy them from the side and have enough time to get the pickup before respawning.

dyceron:
> More general question: How is that supposed to work with the static changes? Should this be pre- or post the static_fixes? Does this even make a difference? The A1 teleporter crumbles doesn't really care what block type they are.
> 
> Can you destroy a crumble block in a one tile gap within a wall or will all those pickups be unacessible then?

I guess it should be done before or not doing the static fixes at all, but then you have the ones in game patches too.

dyceron:
I also forgot about Screw Attack blocks whoops

ThanatosGit (approved):

PR #434, opened 2024-07-27
Add support for custom damage reduction per suit

### LINKED ISSUE #431: Custom Damage Reduction

dyceron:
Fixes #431


ThanatosGit (approved):
Never tried it but I assume it works (not like the scan pulse range :( )

Issue #435, opened 2024-07-28
Caching `get_asset_names_in_folder` causes custom DNA Chozo Seals to not fully load


dyceron:
Only occurs in scenarios without a Chozo Seal where a custom seal has been added. Only the pedestal is loaded and there are memory unread/write errors. Caching adds a performance boost of ~1-2 seconds. Ideally, a new fix should have this caching re-added while also having the seals not have issues.

PR #436, opened 2024-07-28
Revert caching `get_asset_names_in_folder`

dyceron:
The caching causing memory errors in scenarios without a Chozo Seal present, where a custom DNA Seal has been added. This has a side effect of the patcher being ~1-2 seconds slower than before.

Related to #435


ThanatosGit (approved):
Shouldn't be too hard to readd it because imo it could only be a problem if we add new things in between. Just need to make sure to add everything in the beginning.

PR #437, opened 2024-07-28
Add support for configurable required DNA

dyceron:
Changes how DNA is handled as a goal. You still need to collect all DNA like before to access Ridley, but now collecting a configurable amount of DNA is possible instead of collecting all DNA that is placed.
- A new prop has been added to determine if the objective is completed
- The amount of starting DNA is now calculated by taking the difference of the total DNA in the game (39) and the amount of required DNA to finish
- If the amount of required DNA is 0, the prop will be set to true (same as before but now using a prop)
- The DNA text hud now blinks three times after collecting all DNA
- Fixes the DNA text reverting the color back to white if the file is reloaded

https://github.com/user-attachments/assets/b6e3b17d-ab5b-40b4-90ce-fc2fcf32675c


src/open_samus_returns_rando/files/levels/s110_surfaceb.lua
```diff
@@ -560,7 +560,7 @@ function s110_surfaceb.OnSubAreaChange(_ARG_0_, _ARG_1_, _ARG_2_, _ARG_3_, _ARG_
   Game.SetScenarioItemEnabledByName("ray01", false)
   Game.SetSceneGroupEnabledByName("sg_debris02", false)
   Game.SetSceneGroupEnabledByName("sg_debris03", false)
-  if _ARG_2_ == "collision_camera_000" and Game.GetItemAmount(Game.GetPlayerName(), "ITEM_ADN") < 39 then
+  if _ARG_2_ == "collision_camera_000" and Blackboard.GetProp("GAME", "OBJECTIVE_COMPLETE") == false then
```
ThanatosGit, line 563: We usually don't do explicit checks for booleans. Dunno if you want to keep it.

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
```diff
@@ -182,7 +159,7 @@ function RandomizerPowerup.HandlePickupResources(progression, actorOrName)
                         end
                         Scenario.UpdateDNACounter()
                         RandomizerPowerup.IncreaseItemAmount("ITEM_ADN", resource.quantity)
-                        Game.AddSF(0, "RandomizerPowerup.ObjectiveComplete", "")
+                        RandomizerPowerup.ObjectiveComplete()
```
ThanatosGit, line 162: I assume it's fine? Iirc multiworld encapsulates granting an item in a scheduled function anyways, so it shouldn't be a problem.

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
```diff
@@ -217,6 +194,40 @@ function RandomizerPowerup.IncreaseAmmo(resource)
     RandomizerPowerup.IncreaseItemAmount(current_id, resource.quantity, resource.item_id)
 end
 
+function RandomizerPowerup.ObjectiveComplete()
+    if RandomizerPowerup.GetItemAmount("ITEM_ADN") == (39 + Init.iExtraDNA) then
```
ThanatosGit, line 198: Oh. When I was reading the schema, description and names etc. I really thought that "extra" means placed DNA but not required (and was wondering why we would even need that value). Reading this line means, extra DNA must be collected.
Maybe it's just me connecting "extra" always more towards the meaning of something like "bonus".

Dunno how the RDV PR will look like but extra DNA really only makes sense when you want to collect more than 39 DNA.
Which makes me wonder: Do we really still want to count up if the value is not fixed anymore to 39? Because we never show how many DNA you need to collect due to not enough space in the lower bar (which was fine when it always needs to be 39).

src/open_samus_returns_rando/files/levels/s110_surfaceb.lua
dyceron: Ah right. This would normally be `not blah`. I'll fix it.

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
dyceron: I'm pretty sure the reason it was scheduled before was to actually show up after the item popup. But since the popups are asynchronous then it doesn't matter anymore.

src/open_samus_returns_rando/files/templates/randomizerpowerup.lua
dyceron: Honestly, the use case of this is so niche that I might just remove it. If you wanted more dna placed you don't really need this plus this could be checked for a different way.

dyceron: > Dunno how the RDV PR will look like but extra DNA really only makes sense when you want to collect more than 39 DNA. Which makes me wonder: Do we really still want to count up if the value is not fixed anymore to 39? Because we never show how many DNA you need to collect due to not enough space in the lower bar (which was fine when it always needs to be 39).

I would prefer to just have the dna count start at what you need to finish (the required dna) and then count down to 0, but what I tried didn't work and I didn't look too hard, so the behavior is the same as before. Maybe it's trivial to change it and I'm not realizing it.

src/open_samus_returns_rando/files/schema.json
```diff
@@ -255,20 +255,11 @@
...
-                    "type": "number",
-                    "default": 0.5
-                }
-            },
-            "default": {}
+        "required_dna": {
+            "description": "The amount of Metroid DNA required to access Proteus Ridley.",
+            "type": "integer",
+            "minimum": 0,
+            "default": 39
```
ThanatosGit, line 262: Just a comment that this breaks all previous inputs / patcher.json files such that they are not beatable anymore without manual modifications (unless they have all 39 DNA placed).
Don't know if you want to account for that in the next version number.

ThanatosGit (approved):

src/open_samus_returns_rando/files/schema.json
ThanatosGit: Then you give free access to ridley because this will be always true:
https://github.com/randovania/open-samus-returns-rando/blob/f6e76e0251aca6a33cf3d8a0798a23de82a781ef/src/open_samus_returns_rando/files/templates/custom_init.lua#L76-L81

I can't think of any way besides writing a migration like: "If required_dna is not present, count the amount of DNA in starting inventory and set required_dna to 39 - DNA in inventory".
Or just break old files and bump major version.
I'm wondering if anyone is directly using patcher files besides the few people following the development.

Miepee: > I'm wondering if anyone is directly using patcher files besides the few people following the development.

AFAIK RDV is currently the only downstream. So for least effort i'd just bump major+break compat

dyceron: Feels wrong to bump major for something like this. I'd rather restructure the schema if we were to bump major.

Miepee: There's nothing wrong about it and it shouldn't feel wrong either. A new feature that isnt (in a simple way) backwards compatible is a perfectly valid reason for a major bump.

ThanatosGit: > > I'm wondering if anyone is directly using patcher files besides the few people following the development.
> 
> AFAIK RDV is currently the only downstream. So for least effort i'd just bump major+break compat

It's not about RDV. RDV takes a rdvgame and creates a patcher file. Of course it should create the appropriate file once this is merged into RDV even for older rdvgame files. So, RDV already does a migration and makes sure that old rdvgame files do not break.
It's about the fact that with the default you use in the schema, the programm still executes but the output is different such that the game is not beatable anymore, which, at least to me, is a breaking change.
Or you use 0 as default but then the game also changes behaviour because you can directly access ridley.

I would just bump the major version but also it's very likely that nobody will care because there is no one using the patcher directly.

Miepee: > nobody will care because there is no one using the patcher directly.

That's what I meant. The (currently known) only user of OSRR is RDV ,and RDV has migrations in place to not be affected by this.

dyceron: > I can't think of any way besides writing a migration like: "If required_dna is not present, count the amount of DNA in starting inventory and set required_dna to 39 - DNA in inventory".

Honestly, this doesn't sound like a bad idea. Should be easy to add.

PR #441, opened 2024-07-29
Readd cache

### LINKED ISSUE #435: Caching `get_asset_names_in_folder` causes custom DNA Chozo Seals to not fully load

ThanatosGit:
Ok, I lied in the other PR. I just looked into the code. We return a generator via `yield` in the function.
It's a pretty bad idea to use a `cache` on a generator because it will return the same generator multiple times but we iterate through it, so the generator is already exhausted after first loop.
Or short: This code is only executed once and not for all required scenarios.
```
    for asset in editor.get_asset_names_in_folder("actors/props/systemmechdna"):
        editor.ensure_present_in_scenario(scenario_name, asset)
```

I switched to return a list instead.
Should have seen this when I initially reviewed it. Stupid me, sorry. :(

But better to not include for the next release.

Fix #435

### DIFF
--- a/src/open_samus_returns_rando/patcher_editor.py
+++ b/src/open_samus_returns_rando/patcher_editor.py
@@ -1,4 +1,5 @@
 import copy
+import functools
 import typing
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


Miepee:
Can you show the performance change between main / this branch?

dyceron:
Hmm, my testing shows it to be about the same?

ThanatosGit:
Ok

dyceron:
I didn't mean for you to close this. I was hoping you would test this to confirm.

PR #448, opened 2024-08-02
Fix default for `required_dna`

dyceron:


ThanatosGit:
As already discussed: This doesn't fix anything, just changes one problem for a different one.

PR #458, opened 2024-08-18
Schema refactor

dyceron:
This changes the schema so it breaks compatibility with older versions, so a major version bump is necessary.
- `camera_names_dict` has been moved out of cosmetic options and is now called `collision_camera_attributes`.
  - A new field has been added to support adding in custom environmental sources, such as new heat rooms or submerged rooms. (NYI)
  - More fields can be added to this making it flexible and compatible going forward
- `reveal_map_on_start` has been moved to cosmetic patches, though this one is still not very relevant currently

I'm leaving this PR as a draft in case we decide on making other big changes to the schema, or if we possibly do keep compatibility with older versions. I'm open to ideas.

### DIFF
--- a/src/open_samus_returns_rando/files/custom/room_names.lua
+++ b/src/open_samus_returns_rando/files/custom/room_names.lua
@@ -16,7 +16,7 @@ function RoomNameGui.GetRoomName(camera)
         return nil
     end
 
-    return dict[scenario][camera]
+    return dict[scenario][camera]["room_name"]
 end
 
 function RoomNameGui.Init()
--- a/src/open_samus_returns_rando/files/schema.json
+++ b/src/open_samus_returns_rando/files/schema.json
@@ -539,23 +539,6 @@
                     "default": "NEVER",
                     "description": "Configures how the room name is displayed on the HUD. Defaults to disabled."
                 },
-                "camera_names_dict": {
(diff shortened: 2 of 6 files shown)


ThanatosGit (changes requested):
The schema is not fully validating what it should.

dyceron:
Wondering if it's worth the effort to refactor the pickups and doors as well and move them into the cc section, so then we could have something like:
```
    "regions": {
        "s000_surface": {
            "collision_camera_000": {
                "room_name": "Landing Site",
                "pickups": [
                    {
                        "pickup_type": "actor",
                        "caption": "Nothing acquired.",
                        "resources": [
                            [
                                {
                                    "item_id": "ITEM_NONE",
                                    "quantity": 1
                                }
                            ]
                        ],
                        "actor": "LE_Item_005",
                        "model": [
                            "itemsphere"
                        ]
                    }
                ],
                "doors": [    
                    {
                        "actor": "Door012",
                        "door_type": "power_beam"
                    },
                    {
                        "actor": "Door011",
                        "door_type": "power_beam"
                    },
                    {
                        "actor": "Door014",
                        "door_type": "power_bomb"
                    }
                ],
                "custom_env_source": "lava"
            }
        }
    }
```
Could also be good for plandos.

PR #478, opened 2024-10-12
Use rom files directly instead of multiple inputs + change and update patch formats

ThanatosGit:
- requires meds update
- bugfix before release in the code for luma + testing


dyceron (approved):
Lgtm

dyceron:
The root stuff you removed was for reading the romfs directly, right?

dyceron:
Yeah that makes sense

PR #519, opened 2025-01-27
Remove custom `heatzone.lua`

dyceron:
By reverting the heat sound change function in #517, the custom `heatzone.lua` is now a duplicate of the original one and can be removed, as it does nothing.

Also adds a default case to `RandoApi.CheckSuits()` in case we want to be able to remove suits to properly update the suit model.


ThanatosGit:
> Also adds a default case to `RandoApi.CheckSuits()` in case we want to be able to remove suits to properly update the suit model.

Hmm that would require to deal with the liquid stuff, so I don't really see a reason to add this now but also doesn't do any harm.

ThanatosGit (approved):

PR #526, opened 2025-02-05
Have schema determine what sound plays when a pickup is picked up

### LINKED ISSUE #525: Non-Required Mains plays the major item sound for tanks

dyceron:
Fixes #525

### DIFF
--- a/src/open_samus_returns_rando/files/schema.json
+++ b/src/open_samus_returns_rando/files/schema.json
@@ -36,6 +36,15 @@
                     "caption": {
                         "type": "string"
                     },
+                    "sound": {
+                        "type": "string",
+                        "enum": [
+                            "tank_jingle",
+                            "sphere_jingle_placeholder",
+                            "special_ability2_32",
+                            "k_matad_jinchozo"
+                        ]
+                    },
                     "resources": {
                         "type": "array",
                         "items": {
--- a/src/open_samus_returns_rando/lua_editor.py
+++ b/src/open_samus_returns_rando/lua_editor.py
@@ -35,21 +35,6 @@ SPECIFIC_CLASSES = {
     "ITEM_RESERVE_TANK_SPECIAL_ENERGY": "RandomizerReserveTankA",
 }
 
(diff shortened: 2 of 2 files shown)


ThanatosGit:
Uh, was this tested? Because I really think this does nothing.
The sound is determined by the first item id but the first item is always the main item because non-required mains is exported as every item gives you the main item.
It should be e.g. a list of two items where the first item is `ITEM_WEAPON_POWER_BOMB`  and the second item is `ITEM_POWER_BOMB_TANKS` (so, _MAX is not even used anywhere because we have the `export_item_id` extra in randovania)

ThanatosGit:
Small addendum: Of course, a possible option is to export non-required mains differently from randovania by only using the `_MAX` item id. In that case this PR would be the first step but is not fixing the issue.
No idea how hacky this is gonna be in randovania but if you want to make these changes, then the PR itself is fine.

dyceron:
I'll revisit this pr.

dyceron:
Yeah it turns out this doesn't do anything. I'm not sure the best approach to this.

ThanatosGit:
Hmmm I'm not a big fan of relying on arbitrary captions for the jingles.
(For whatever reason) renaming the caption from e.g. Tank to Expansion in randovania would break the jingles in the patcher which is not really intuitive.
There are also - I must admit not very important - corner cases: If I have a multi world or even better: A coop world and have anywhere the string "Tank" in it like "blabla blaTankbla" it would export all captions like `Sent Morph Ball to blabla blaTankbla`  and matches always the very first if clause and everything would use the same jingle.
I think it's just a bad design and logically we don't export something which clearly tells us stuff like "it is a major item but it uses a hidden model ", "this one uses a hidden caption thing" etc. 
That context is only known in randovania and I begin to think that it would be a better design to add a jingle to the patcher schema and leave it to the pdf to decide which one to use.

dyceron:
I changed it so the pickup sounds are handled by the schema. This makes the most sense to me, and is easily customizable.

ThanatosGit (approved):

PR #528, opened 2025-02-06
Display number of DNA left in boss message popup

dyceron:
Also cleans up some minor related things

image

### DIFF
--- a/src/open_samus_returns_rando/files/levels/s070_area7.lua
+++ b/src/open_samus_returns_rando/files/levels/s070_area7.lua
@@ -166,7 +166,9 @@ end
 function s070_area7.OnEnter_Diggernaut_Access()
   if Init.sFinalBoss == "Diggernaut" then
     if not Blackboard.GetProp("GAME", "OBJECTIVE_COMPLETE") then
-      Game.AddSF(0, "Scenario.ShowFinalBossMessage", "")
+       local startpoint = "Diggernaut" 
+       GUI.LaunchMessage("Not enough Metroid DNA!\nCollect " .. RandomizerPowerup.GetItemAmount(ITEM_ADN) .. " more DNA to fight Diggernaut!", "RandomizerPowerup.Dummy", "") 
+       Game.AddSF(0, "Scenario.FinalBossReload", "s", startpoint)
     else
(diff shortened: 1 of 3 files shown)


src/open_samus_returns_rando/files/levels/s070_area7.lua
```diff
@@ -166,9 +166,7 @@ end
 function s070_area7.OnEnter_Diggernaut_Access()
   if Init.sFinalBoss == "Diggernaut" then
     if not Blackboard.GetProp("GAME", "OBJECTIVE_COMPLETE") then
-      local startpoint = "Diggernaut"
```
ThanatosGit, line 169: This is changing the startpoint from `Diggernaut` to `ManicMiner` because that's what is used here: https://github.com/randovania/open-samus-returns-rando/blob/578b80a25ad26d60be0810817472ee132767565a/src/open_samus_returns_rando/files/custom/scenario.lua#L210-L211

Intentional? I don't know what is the correct one

dyceron: Definitely not what I meant to do. This should be reverted.

ThanatosGit: I can't really follow now. Why was it reverted instead of just changing the string from the quoted lines above?
When do you need to use the startpoint `ManicMiner` and when `Diggernaut`?

dyceron: iirc I named it differently to differentiate from the vanilla ManicMiner startpoint, since there are two spawns. Idk it probably could be changed but this already works.

ThanatosGit: If it is never used in `Scenario.ShowFinalBossMessage()`, we can change it and don't need to have the message in two places that need to be kept synchronised. blabla clean code blabla dry principle blabla

dyceron: I'm fine with cleaning it up, but it could also be done as a later pr. Up to you.

ThanatosGit (approved):

PR #532, opened 2025-02-14
Fix configuration default for `final_boss_hint`

dyceron:
~Now properly gets the default value if not present. Before it was saying `configuration[configuration["final_boss_hint"]]`~

Now ensures that a message is provided regardless,

### DIFF
--- a/src/open_samus_returns_rando/lua_editor.py
+++ b/src/open_samus_returns_rando/lua_editor.py
@@ -268,7 +268,7 @@ class LuaEditor:
         # Handle older versions of the patcher
         if "baby_metroid_hint" in configuration or "final_boss_hint" in configuration:
             final_boss_hint = lua_util.wrap_string(
-                configuration.get("baby_metroid_hint", configuration["final_boss_hint"])
+                configuration.get("baby_metroid_hint", "final_boss_hint")
             )
 
         if "required_dna" in objective:
--- a/src/open_samus_returns_rando/lua_editor.py
+++ b/src/open_samus_returns_rando/lua_editor.py
@@ -263,13 +263,12 @@ class LuaEditor:
         if "layout_uuid" in configuration:
             layout_uuid = lua_util.wrap_string(configuration["layout_uuid"])
 
-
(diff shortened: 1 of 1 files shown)


duncathan:
... are you sure this is the behavior you want?

dyceron:
> ... are you sure this is the behavior you want?

Oops...

ThanatosGit:
What will be fixed with this PR? Logically, the old and the new block look identical to me. The newer block is shorter and probably better to read but I'm wondering about the `Fix` part in this PR and if it is missing something.

The old block is:
```python

        final_boss_hint = lua_util.wrap_string("No hint for the final boss was provided!")
        # Handle older versions of the patcher
        if "baby_metroid_hint" in configuration or "final_boss_hint" in configuration:
            final_boss_hint = lua_util.wrap_string(
                configuration.get("baby_metroid_hint", configuration["final_boss_hint"])
            )
```
It starts with setting the `final_boss_hint ` to `"No hint for the final boss was provided!"`, then checks if `baby_metroid_hint` is in `configuration` or if `final_boss_hint` is in `configuration` . 
Iff that's true, it prefers to use the value of `baby_metroid_hint` in `configuration`, otherwise it uses the value of `final_boss_hint` in `configuration`
Iff false, it does nothing and keeps the initial value of `"No hint for the final boss was provided!"`.

New block:
```python
        # Handle older versions of the patcher
        final_boss_hint = lua_util.wrap_string(
            configuration.get(
                "baby_metroid_hint", configuration.get("final_boss_hint", "No hint for the final boss was provided!")
            )
        )
```

Tries to get the value of `baby_metroid_hint` in `configuration`, iff not present it tries to get value of `final_boss_hint` in configuration, iff not present uses  ` "No hint for the final boss was provided!"`.

dyceron:
With the old block, if I had `baby_metroid_hint` in the config, it would throw a key error saying about `final_boss_hint`. I don't have this issue with the new block.

ThanatosGit:
Oh, lol, yeah makes sense because the second parameter of the get access it directly.
I will stop reading code in the early morning 🙃

ThanatosGit (approved):

PR #533, opened 2025-02-14
Fix Reserve Tank FX not showing if it's for another world

### LINKED ISSUE #531: Reserve Tanks for other players don't display fx

dyceron:
Hacky fix but it works. Probably should add custom models at some point but for now this gets the job done.

Fixes #531

### DIFF
--- a/src/open_samus_returns_rando/pickups/pickup.py
+++ b/src/open_samus_returns_rando/pickups/pickup.py
@@ -14,10 +14,13 @@ from open_samus_returns_rando.lua_editor import LuaEditor, ScriptClass
 from open_samus_returns_rando.patcher_editor import PatcherEditor, path_for_level
 from open_samus_returns_rando.pickups.model_data import get_data
 
-RESERVE_TANK_ITEMS = {
+RESERVE_TANKS = {
     "ITEM_RESERVE_TANK_LIFE",
     "ITEM_RESERVE_TANK_MISSILE",
     "ITEM_RESERVE_TANK_SPECIAL_ENERGY",
+    "powerup_reserveenergy",
+    "powerup_reservemissile",
+    "powerup_reserveaeion",
 }
 
 TANK_MODELS = {
@@ -151,8 +154,8 @@ class ActorPickup(BasePickup):
                 else:
                     MODELUPDATER["functions"][0]["params"].pop("Param2")
                 # Placeholder until custom models/textures are made
-                # Checks for caption in multiworld if for another world
-                if item_id in RESERVE_TANK_ITEMS or "Reserve Tank" in self.pickup["caption"]:
+                # FIXME: Make this less ugly
+                if model_name in RESERVE_TANKS or item_id in RESERVE_TANKS:
                     fx_create_and_link["Param1"]["value"] = "spinattack"
                     fx_create_and_link["Param2"]["value"] = "actors/characters/samus/fx/spinattack.bcptl"
                     fx_create_and_link["Param8"]["value"] = 55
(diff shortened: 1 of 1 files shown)


src/open_samus_returns_rando/pickups/pickup.py
```diff
@@ -151,7 +151,8 @@ def patch_model(self, model_names: list[str], bmsad: dict) -> None:
                 else:
                     MODELUPDATER["functions"][0]["params"].pop("Param2")
                 # Placeholder until custom models/textures are made
-                if item_id in RESERVE_TANK_ITEMS:
+                # Checks for caption in multiworld if for another world
+                if item_id in RESERVE_TANK_ITEMS or "Reserve Tank" in self.pickup["caption"]:
```
ThanatosGit, line 155: I don't like that for the same reason as in the other PR: The caption string is not a thing I want to parse to determine items, models or whatsoever.

I would suggest to just add a new intermediate model name like `item_reserve_missiletank` which gets "translated" by the patcher to `item_missiletank` + the FX. Should be a very small change in randovania (one line or three lines for all reserve tanks?) and whenever you really want  to use a fully custom model or similiar, you don't need to change anything in randovania after that.
Also shouldn't be too complicated here in the patcher, no? And we still have our `item_models_test.json` which makes it easy to test the changes.

Tbh I think the whole `ModelData` thing should be refactored to contain the FX instead of doing plenty of our messy `if model_name in blabla` but that's probably a too long stretch for this PR.

ThanatosGit (changes requested):

src/open_samus_returns_rando/pickups/pickup.py
dyceron: The patcher as a whole could get a major refactor honestly. But that's for later.

src/open_samus_returns_rando/pickups/pickup.py
```diff
@@ -151,8 +154,8 @@ def patch_model(self, model_names: list[str], bmsad: dict) -> None:
                 else:
                     MODELUPDATER["functions"][0]["params"].pop("Param2")
                 # Placeholder until custom models/textures are made
-                # Checks for caption in multiworld if for another world
-                if item_id in RESERVE_TANK_ITEMS or "Reserve Tank" in self.pickup["caption"]:
+                # FIXME: Make this less ugly
+                if model_name in RESERVE_TANKS or item_id in RESERVE_TANKS:
```
ThanatosGit, line 158: I thought about asking to remove the `item_id` part but then I realised that's for compatibility with old patcher files, right?

ThanatosGit (approved):

src/open_samus_returns_rando/pickups/pickup.py
```diff
@@ -14,10 +14,13 @@
...
 from open_samus_returns_rando.pickups.model_data import get_data
 
-RESERVE_TANK_ITEMS = {
+RESERVE_TANKS = {
     "ITEM_RESERVE_TANK_LIFE",
     "ITEM_RESERVE_TANK_MISSILE",
     "ITEM_RESERVE_TANK_SPECIAL_ENERGY",
+    "powerup_reserveenergy",
+    "powerup_reservemissile",
+    "powerup_reserveaeion",
```
ThanatosGit, line 23: Oh wait: Doesn't these new values need entries in the `MODEL_DATA` thing for the bcmdl path and dependencies etc.?

src/open_samus_returns_rando/pickups/pickup.py
dyceron: Yes but idk how important that is. If we ended up making a completely new model, then it'd be wrong anyway.

dyceron: Actually this isn't even going to work at all, because first it's checking if it's a tank model, then checking for the new model names. So the first check fails. I hate this part of the pickup code lol

dyceron:
Closing in favor of a better future approach

PR #534, opened 2025-02-15
Validate starting actor exists

### LINKED ISSUE #530: Throw error if starting_actor is not valid
Setting an invalid spawn causes Samus to spawn in a predefined location (no idea if it's a centered location on the map). Checking to see if the actor is valid is preferable to prevent unintended results.

dyceron:
Fixes #530

### DIFF
--- a/src/open_samus_returns_rando/lua_editor.py
+++ b/src/open_samus_returns_rando/lua_editor.py
@@ -212,6 +212,15 @@ class LuaEditor:
         configuration_identifier: str = configuration["configuration_identifier"]
         enable_remote_lua: bool = configuration.get("enable_remote_lua", False)
 
+        starting_scenario = starting_location["scenario"]
+        scenario_bmsld = editor.get_scenario(starting_scenario)
+        starting_actor = starting_location["actor"]
+        for actor in scenario_bmsld.raw.actors[5]:
+            if starting_location["actor"] == actor:
+                break
+        else:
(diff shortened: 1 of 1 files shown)


src/open_samus_returns_rando/lua_editor.py
```diff
@@ -212,6 +212,15 @@ def _create_custom_init(self, editor: PatcherEditor, configuration: dict) -> str
         configuration_identifier: str = configuration["configuration_identifier"]
         enable_remote_lua: bool = configuration.get("enable_remote_lua", False)
 
+        starting_scenario = starting_location["scenario"]
+        scenario_bmsld = editor.get_scenario(starting_scenario)
+        starting_actor = starting_location["actor"]
+        for actor in scenario_bmsld.raw.actors[5]:
+            if starting_location["actor"] == actor:
+                break
+        else:
```
Miepee, line 221: i'm not thanatos, but i would prefer to have a temp variable instead of using for...else.

dyceron: How would this help?

Miepee: more readable for me. I find for-else to be obscure syntax.

dyceron: I just updated it. Definitely cleaner.

ThanatosGit (approved):

PR #536, opened 2025-02-25
Refactor pickups

### LINKED ISSUE #531: Reserve Tanks for other players don't display fx

dyceron:
Quite a lot to unpack here, so here goes:
- The minimap icon is now handled by the schema instead of being hardcoded
  - This should fix hidden item models funnily enough
- `model_data.py` now also tracks transforms, fx, and action_sets
- "New" models were added for the Reserve Tanks
  - These are the same models/fx as before, but now they are part of `model_data.py`. Currently, this just converts the tank model to the new model. In the event we actually do make new models, we already have something to work with.
- `template_powerup_bmsad.json` was updated to include common data and also to remove extra work from `pickup.py`
- Overall cleanup of `pickup.py` to a more readable state
  - There were a bunch of defined fields before like sounds, popping certain fields, etc. These were all removed, as they were irrelevant to begin with.

There's definitely more that could be done here, and possibly some issues I did not encounter in my testing, but this is much better than it was before.

Fixes #531


ThanatosGit (approved):
The code is now too easy to read. Can you add some `if` and `elif` here and there?

PR #545, opened 2025-04-23
Add mapping from model name to icon

ThanatosGit:


src/open_samus_returns_rando/pickups/pickup.py
```diff
@@ -12,14 +12,9 @@
...
 from open_samus_returns_rando.patcher_editor import PatcherEditor, path_for_level
+from open_samus_returns_rando.pickups.map_icons import get_map_icon_data
 from open_samus_returns_rando.pickups.model_data import get_data
 
-RESERVE_TANK_ITEMS = {
-    "ITEM_RESERVE_TANK_LIFE": "powerup_energyreservetank",
-    "ITEM_RESERVE_TANK_SPECIAL_ENERGY": "powerup_aeionreservetank",
-    "ITEM_RESERVE_TANK_MISSILE": "powerup_missilereservetank",
-}
-
```
ThanatosGit, line 22: This breaks old patcher files, but I have no idea why we had a mapping for the item id to a model instead of using the correct model in rdv in the first place

ThanatosGit: Oh wait, was this actually every released that was?

dyceron: They were just assigned the tank model, so I had to use the id to determine if it should have the fx or not. Not the best solution. It's been in releases yes. Don't know how much we care about breaking older patcher for this. I personally don't care but eh.

src/open_samus_returns_rando/pickups/pickup.py
```diff
@@ -226,7 +217,10 @@ def patch_minimap(self, editor: PatcherEditor, scenario_name: str, actor_name: s
                 else:
                     pickup_tile_icon.icon = "itemenabled"
             else:
-                pickup_tile_icon.icon = self.pickup["map_icon"]
+                first_model_name = self.pickup["model"][0]
+                icon = self.pickup.get("map_icon", None)
+                icon = get_map_icon_data(first_model_name) if icon is None else icon
```
dyceron, line 222: I assume this reassignment is fine. It just looks weird but I understand what's going on.

src/open_samus_returns_rando/pickups/map_icons.py
```diff
@@ -0,0 +1,26 @@
...
+    "item_powerbombtank": "item_powerbombtank",
+    "item_senergytank": "item_senergytank",
+
+    # reserve tanks
+    "powerup_missilereservetank": "item_missiletank",
+    "powerup_energyreservetank": "item_energytank",
+    "powerup_aeionreservetank": "item_senergytank",
+
+    # offworld
+    "offworld_generic": "item_offworld",
```
dyceron, line 18: Missing the other offworld models

dyceron (changes requested):

src/open_samus_returns_rando/pickups/pickup.py
ThanatosGit: :(
In retrospect this should have just been a new model name like it is now.
Iirc so far we always bumped the major version if we had breaking changes for old input files. Then I revert that part.

src/open_samus_returns_rando/pickups/pickup.py
ThanatosGit: I saved a total of 1 line with that!
...I changed it :p

dyceron (approved):

PR #552, opened 2025-07-06
Account for varying starting/objective DNA amounts

dyceron:
Fixes https://discord.com/channels/914291389293027329/1386412573498282024 

Depending on the settings used, DNA could be given as random starting items with no other DNA placed in the world. This should account for all cases of different combinations of starting DNA and required DNA.

### DIFF
--- a/src/open_samus_returns_rando/lua_editor.py
+++ b/src/open_samus_returns_rando/lua_editor.py
@@ -279,12 +279,15 @@ class LuaEditor:
 
         starting_dna = [item for item in inventory if item.startswith("ITEM_RANDO_DNA")]
         default_dna = 39
-        objective_dna = objective["required_dna"]
         if "required_dna" in objective:
+            objective_dna = objective["required_dna"]
+            # If the objective is higher than default, take the difference of objective and starting
             if objective_dna >= default_dna:
                 required_dna = objective_dna - len(starting_dna)
+            # If the objective is less than default and starting equals default, require 0 dna
             elif len(starting_dna) == default_dna:
                 required_dna = 0
+            # If the objective is less than default, require objective
             else:
                 required_dna = objective_dna
         else:
--- a/src/open_samus_returns_rando/lua_editor.py
+++ b/src/open_samus_returns_rando/lua_editor.py
@@ -279,7 +279,7 @@ class LuaEditor:
 
         starting_dna = [item for item in inventory if item.startswith("ITEM_RANDO_DNA")]
         total_dna = objective.get("total_dna", 39)
-        placed_dna = objective.get("placed_dna", 0)
+        placed_dna = objective.get("placed_dna", -1 * (total_dna - len(starting_dna))
         objective_dna = objective.get("required_dna", 0)
 
         # Handle starting with extra DNA which would affect the objective
--- a/src/open_samus_returns_rando/lua_editor.py
+++ b/src/open_samus_returns_rando/lua_editor.py
@@ -279,7 +279,7 @@ class LuaEditor:
 
         starting_dna = [item for item in inventory if item.startswith("ITEM_RANDO_DNA")]
         total_dna = objective.get("total_dna", 39)
-        placed_dna = objective.get("placed_dna", total_dna - len(starting_dna)
+        placed_dna = objective.get("placed_dna", total_dna - len(starting_dna))
         objective_dna = objective.get("required_dna", 0)
 
         # Handle starting with extra DNA which would affect the objective
--- a/src/open_samus_returns_rando/lua_editor.py
+++ b/src/open_samus_returns_rando/lua_editor.py
@@ -278,20 +278,18 @@ class LuaEditor:
         )
 
         starting_dna = [item for item in inventory if item.startswith("ITEM_RANDO_DNA")]
-        default_dna = 39
-        if "required_dna" in objective:
-            objective_dna = objective["required_dna"]
-            # If the objective is higher than default, take the difference of objective and starting
-            if objective_dna >= default_dna:
-                required_dna = objective_dna - len(starting_dna)
-            # If the objective is less than default and starting equals default, require 0 dna
-            elif len(starting_dna) == default_dna:
-                required_dna = 0
-            # If the objective is less than default, require objective
-            else:
-                required_dna = objective_dna
+        total_dna = objective.get("total_dna", 39)
+        objective_dna = objective.get("required_dna", 0)
+        placed_dna = total_dna - len(starting_dna)
+        # If there is no placed DNA or there is no objective DNA, require 0 DNA
+        if placed_dna == 0 or objective_dna == 0:
+            required_dna = 0
+        # If there are less DNA placed than required, require DNA that is placed
+        elif placed_dna < objective_dna:
+            required_dna = placed_dna
+        # If the same amount or more DNA is placed than required, require DNA from the objective
         else:
-            required_dna = default_dna - len(starting_dna)
+            required_dna = objective_dna
 
         replacement = {
             "new_game_inventory": final_inventory,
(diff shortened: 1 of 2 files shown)


src/open_samus_returns_rando/lua_editor.py
```diff
@@ -277,11 +277,21 @@ def chunks(array: list[str], n: int) -> Iterable[list[str]]:
...
+            objective_dna = objective["required_dna"]
+            # If the objective is higher than default, take the difference of objective and starting
+            if objective_dna >= default_dna:
+                required_dna = objective_dna - len(starting_dna)
+            # If the objective is less than default and starting equals default, require 0 dna
+            elif len(starting_dna) == default_dna:
+                required_dna = 0
+            # If the objective is less than default, require objective
+            else:
+                required_dna = objective_dna
```
ThanatosGit, line 292: That case happens if `objective_dna < 39` but also not all DNAs are in the starting inventory. That's likely the default case of 10 DNAs are required and shuffled.
I was writing all this because I think I'm missing something: Can't you have an `objective_dna = 10` but one of those was also placed additionally into the starting inventory such that you still need to reduce the value `objective_dna` by the (additional) starting DNA? Is it not possible from RDV?

Dunno if I'm making up super weird edge cases that don't exist like: What happens if the preset in RDV says "Shuffle 12, require only 6 of them" but then one of those shuffled DNAs is also added as starting item? (once again: Don't know if this is actually possible). Then in the patcher file and code we have `objective_dna = 6` and `starting_dna = 28` (`39 - 12 (shuffled) + 1 (added as random starting item)`). I can't even see a way to detect this in our current patcher json but you would need to reduce the `required_dna` in that case by `1` => `required_dna = 6 (objective_dna) - 1 (added as random starting item)`

ThanatosGit: Why it has to be all of the extra DNA as starting items?
Based on my example, I thought it is sufficient to only have one of the 12 as starting item.
The `objective_dna` value is the one from the patcher.json which is 6. That value comes from the preset settings. 12 should be shuffled according to the preset.
If we don't subtract the additional starting DNA item from the `required_dna` variable, the player still needs to collect 6 DNA but only 11 are shuffled now (because one of them was "placed" as starting item).

dyceron: > Does logic guarantees that all Artifacts are accessible even if not required?

Yes

dyceron: I had considered adding a new field in the objective that is the total dna that is supposed to be in the seed, regardless if starting or not. This could make calculating the required DNA based on starting and comparing it to the total much simpler.

ThanatosGit: > I had considered adding a new field in the objective that is the total dna that is supposed to be in the seed

Is this just to confirm what I wrote? Because I cannot see a difference to "total dna that is supposed to be in the seed" and "amount of shuffled dna" I used for the calculation above.

dyceron: Essentially yeah. So then we could have a clear idea of the actual required DNA, and if the goal were more than 39, it could easily be reflected. Ill add this later.

src/open_samus_returns_rando/lua_editor.py
```diff
@@ -277,11 +277,19 @@ def chunks(array: list[str], n: int) -> Iterable[list[str]]:
             )
         )
 
-        if "required_dna" in objective:
-            required_dna = objective["required_dna"]
+        starting_dna = [item for item in inventory if item.startswith("ITEM_RANDO_DNA")]
+        total_dna = objective.get("total_dna", 39)
+        objective_dna = objective.get("required_dna", 0)
+        placed_dna = total_dna - len(starting_dna)
```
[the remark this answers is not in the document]
ThanatosGit: Well, that's where we disagree. If I make a 6/12 seed, I want to have 6 "spare" DNAs. If I get one of them through random starting items, this should also reduce the amount of DNA to collect. Essentially, this becomes a 5/11 DNA seed, so I still have 6 spare DNAs. Or in other words: It's like playing a 6/12 DNA seed and immediately collect one of them => 5/11.

Just reducing the spare DNAs makes the seed (perhaps) unintentionally more difficult than the player thought. In your extreme case, suddenly all DNAs need to be collected and there are no spare DNAs at all.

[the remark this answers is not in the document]
dyceron: Regardless, starting with more dna is a small edge case already, but we should decide how we want to handle this. Either accept that there will be less dna to collect, or prevent starting with more dna than set.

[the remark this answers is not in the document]
ThanatosGit: Instead of the new `total_dna` field, add a `shuffled_dna` field or better to say: Add both "slider" values in the patcher.json from the preset here:
image

Then the math is:
```
additional_dna = shuffled_dna - (39 - starting_dna)
required_dna = objective_dna - additional_dna 
```
Then a 6/12 seed becomes a 5/11 seed with one random starting DNA.

Unless you really want to disallow a 6/12 seeds to become a 5/11 seed with my solution or a 6/11 seed with your solution, than you have to disallow it somehow in rdv (like you said), yeah.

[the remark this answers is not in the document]
dyceron: Went with your formula idea. I believe this should be fine

src/open_samus_returns_rando/lua_editor.py
```diff
@@ -277,11 +277,14 @@ def chunks(array: list[str], n: int) -> Iterable[list[str]]:
...
-            starting_dna = [item for item in inventory if item.startswith("ITEM_RANDO_DNA")]
-            required_dna = 39 - len(starting_dna)
+        starting_dna = [item for item in inventory if item.startswith("ITEM_RANDO_DNA")]
+        total_dna = objective.get("total_dna", 39)
+        placed_dna = objective.get("placed_dna", 0)
+        objective_dna = objective.get("required_dna", 0)
+
+        # Handle starting with extra DNA which would affect the objective
+        additional_dna = placed_dna - (total_dna - len(starting_dna))
+        required_dna = objective_dna - additional_dna
```
ThanatosGit, line 287: I don't want to be a annoying and probably nobody cares about the backward compatibility of patcher.json files but
```
placed_dna = objective.get("placed_dna", 0)
```
is a bad default choice.  That would lead to `additional_dna = 0 - (39 - len(starting_dna))`, so it will be negative and the next line will raise the `required_dna` to a value higher than the collectible DNA.
We cannot fix old patcher.json files because we require the new infomraiton to fix it but to not break the old working patcher.json files without random starting DNA, we could force `additional_dna` to be 0.
E.g. by if-else logic or by make it mathematical always be zero as default `placed_dna = objective.get("placed_dna", -1 * (total_dna - len(starting_dna))`

dyceron: That's fine. Ill make the change

ThanatosGit: Brain fart by me, but luckily you copied the missing closing `)` and the tests fail. It must be
```
placed_dna = objective.get("placed_dna", total_dna - len(starting_dna))
```

dyceron: Fixed

src/open_samus_returns_rando/lua_editor.py
```diff
@@ -277,11 +277,14 @@ def chunks(array: list[str], n: int) -> Iterable[list[str]]:
...
         )
 
-        if "required_dna" in objective:
-            required_dna = objective["required_dna"]
-        else:
-            starting_dna = [item for item in inventory if item.startswith("ITEM_RANDO_DNA")]
-            required_dna = 39 - len(starting_dna)
+        starting_dna = [item for item in inventory if item.startswith("ITEM_RANDO_DNA")]
+        total_dna = objective.get("total_dna", 39)
+        placed_dna = objective.get("placed_dna", total_dna - len(starting_dna)
```
ThanatosGit, line 282: still missing a `)`

dyceron: Oops

ThanatosGit (approved):

PR #564, opened 2025-10-08
Add an "item" as a check for completing the game (for RDV)

dyceron:
Relevant for https://github.com/randovania/randovania/pull/8502

I want to say this is all that is needed to make this work. I'm not entirely sure if I need to include this in the schema like I did. Also, I didn't create a new item in the traditional way like the other custom items, but it shouldn't be necessary since you never really collect it. It'd be nicer if RDV could just read the blackboard somehow but this will do for now.


Miepee (commented):
I know the rdv connector does read some stuff from the blackboard. I've been meaning to look into it to see if we can rework the game status stuff to also send whether we've reached credits

PR #577, opened 2026-02-18
Fix for when relative paths are used for the output

### LINKED ISSUE #576: Patcher creates a duplicate file tree if supplied a relative path
Passing a relative path into `samus_returns_patcher.patch_extracted()` causes it to duplicate the relative path into the romfs, which at minimum breaks pickup models.

File tree showing a duplicated nesting

dyceron:
Also fixes readme error

Fixes #576

### DIFF
--- a/src/open_samus_returns_rando/cli.py
+++ b/src/open_samus_returns_rando/cli.py
@@ -58,16 +58,13 @@ def main() -> None:
     args = parser.parse_args()
     print(args)
 
-    if not args.output_path.is_absolute():
-        raise ValueError(f"'{args.output_path}' is not a valid output path. Paths must be absolute!")
-
     with args.input_json.open() as f:
         configuration = json.load(f)
 
     start = time.time()
     samus_returns_patcher.patch_extracted(
         args.input_path,
-        args.output_path,
+        args.output_path.resolve(),
         configuration,
     )
     end = time.time()
--- a/src/open_samus_returns_rando/cli.py
+++ b/src/open_samus_returns_rando/cli.py
@@ -58,6 +58,9 @@ def main() -> None:
     args = parser.parse_args()
     print(args)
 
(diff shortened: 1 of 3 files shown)


src/open_samus_returns_rando/cli.py
```diff
@@ -58,6 +58,9 @@ def main() -> None:
     args = parser.parse_args()
     print(args)
 
+    if not args.output_path.is_absolute():
+        raise ValueError(f"'{args.output_path}' is not a valid output path. Paths must be absolute!")
```
Miepee: instead of erroring, why not resolve the relative path?
```
>>> from pathlib import Path
>>> path = Path(".")
>>> path
PosixPath('.')
>>> path = path.resolve()
>>> path
PosixPath('/home/narr')
>>> 
```
its very strange IMO to error completely out relatives

Miepee: especially as just erroring out breaks our tests :)

dyceron: I wasn't aware of this. Thanks.

src/open_samus_returns_rando/samus_returns_patcher.py
```diff
@@ -150,11 +150,14 @@ def patch_extracted(input_path: Path, output_path: Path, configuration: dict) ->
         # Patch actor attributes
         patch_actor_attributes(editor, configuration["actor_attributes"])
 
-        cleanup_old_patcher(output_path)
+        # Resolve relavtive paths to prevent breaking RomFS folder
```
Miepee: I'd personally use

```suggestion
        # FIXME: Using relative paths breaks the RomFS folder
```

to indicate that resolving the path is moreso a bandaid than the actual solution.

Miepee (approved):

PR #578, opened 2026-03-01
Add README files and submodules for more streamlined IPS patches

ThanatosGit:
For usage see the README files but stripped down:
1. Clone this Repo
2. Init the submodules `git submodule update --init --recursive`
3. Create a venv, activate it and install all dependencies
4. Additonally install the requirements from `external/magikoopa-py`
5. Install devKitPro for windows or linux. See https://devkitpro.org/wiki/Getting_Started
6. Copy unmodified `code.bin` and `exheader.bin` to `external/msr-remote-connector/`

Then you only run `python .\external\build_patches.py` to create the new binary files and create the IPS patch files whenever you want to recreate it

magikoopa-py was generated by an LLM. I just compared the output from the C++/QT version to the python version. It was matching on linux.
The version from windows magikoopa-py did not match the version from linux but I assume that is because of different arm-none-eabi-* compilers. On windows it created an additonal offset of 16 bytes on the first jump instructions to the loader but the loader is also placed 16 bytes later in the binary. So I had no issues on Azahar with it.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -16,7 +16,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           fetch-depth: 0
           submodules: 'recursive'
@@ -36,7 +36,7 @@ jobs:
         run: python -m build --sdist
       
       - name: Store the packages
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v6
         with:
           name: python-package-distributions
           path: dist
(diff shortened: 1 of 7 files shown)


dyceron (approved):
I feel like I understand what's going on here. Would it make sense to include `magikoopa-py` in `pyproject.toml` as a dependency?

ThanatosGit:
> Would it make sense to include `magikoopa-py` in `pyproject.toml` as a dependency?

It's only required to create the IPS patches we directly commit to the patcher. It is just a tool used and not a dependency and also very important: `magikoopa-py` is not even a published package at pypi.

PR #591, opened 2026-04-05
Update to MEDS 0.36.2

ThanatosGit:
Added a commit to the existing branch. Required way more changes than I thought because of the changes to `Vec3` and enums.

### DIFF
--- a/src/open_samus_returns_rando/specific_patches/cosmetic_patches.py
+++ b/src/open_samus_returns_rando/specific_patches/cosmetic_patches.py
@@ -1,3 +1,4 @@
+from mercury_engine_data_structures.common_types import Vec3
 from mercury_engine_data_structures.formats import Bmdefs, Bmses, Bmtun
 
 from open_samus_returns_rando.constants import ALL_SCENARIOS
@@ -14,10 +15,10 @@ def patch_cosmetics(editor: PatcherEditor, configuration: dict) -> None:
 
 def tunable_cosmetics(tunables: Bmtun, configuration: dict) -> None:
     aim = tunables.raw["classes"]["CTunableAim"]["tunables"]
-    aim["vLaserLockedColor0"]["value"] = configuration["laser_locked_color"]
-    aim["vLaserUnlockedColor0"]["value"] = configuration["laser_unlocked_color"]
-    aim["vGrappleLaserLockedColor0"]["value"] = configuration["grapple_laser_locked_color"]
-    aim["vGrappleLaserUnlockedColor0"]["value"] = configuration["grapple_laser_unlocked_color"]
+    aim["vLaserLockedColor0"]["value"] = Vec3(*configuration["laser_locked_color"])
+    aim["vLaserUnlockedColor0"]["value"] = Vec3(*configuration["laser_unlocked_color"])
+    aim["vGrappleLaserLockedColor0"]["value"] = Vec3(*configuration["grapple_laser_locked_color"])
+    aim["vGrappleLaserUnlockedColor0"]["value"] = Vec3(*configuration["grapple_laser_unlocked_color"])
 
 
 def lua_cosmetics(configuration: dict) -> str:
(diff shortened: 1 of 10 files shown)


src/open_samus_returns_rando/specific_patches/cosmetic_patches.py
dyceron, line 1: Not going to use the new function for handling setting tunables?

ThanatosGit: Oh, if we have one we should use it.

dyceron: Yes we do

ThanatosGit: added a commit

src/open_samus_returns_rando/specific_patches/cosmetic_patches.py
```diff
@@ -6,18 +7,20 @@
...
 
 
-def tunable_cosmetics(tunables: Bmtun, configuration: dict) -> None:
-    aim = tunables.raw["classes"]["CTunableAim"]["tunables"]
-    aim["vLaserLockedColor0"]["value"] = configuration["laser_locked_color"]
-    aim["vLaserUnlockedColor0"]["value"] = configuration["laser_unlocked_color"]
-    aim["vGrappleLaserLockedColor0"]["value"] = configuration["grapple_laser_locked_color"]
-    aim["vGrappleLaserUnlockedColor0"]["value"] = configuration["grapple_laser_unlocked_color"]
+def tunable_cosmetics(editor: PatcherEditor, configuration: dict) -> None:
+    # FIXME: Bug in MEDS: It does not accept a list[float] only a Vec3
```
dyceron: I assume this will be fixed in a future MEDS pr?

dyceron (commented):
Is the goal of this pr to use all the new functions or just fix the broken stuff?

ThanatosGit:
> Is the goal of this pr to use all the new functions or just fix the broken stuff?

Both. Do you know any other new methods in 0.36.0 we should use?

src/open_samus_returns_rando/specific_patches/cosmetic_patches.py
ThanatosGit: Already done. https://github.com/randovania/mercury-engine-data-structures/pull/322

dyceron:
> > Is the goal of this pr to use all the new functions or just fix the broken stuff?
> 
> 
> 
> Both. Do you know any other new methods in 0.36.0 we should use?

We have methods for BMSBK. Might be another one I'm forgetting.

ThanatosGit:
Ok, added. That really needs to be double checked and should be tested in-game before a new release.

ThanatosGit:
I byte compared the output of the current released version and this state of the branch. I added a fix for hidden tile icons because they were wrong.

Now, there is only one difference in s000_surface_bmsbk and in the s000_surface_discardables.pkg but as the bmsbk is in the discardables.pkg, it really is just the bmsbk file:

Left is the current released verison and right the pr. It looks like we have this wrong in the released version to me. Because the block should be with the underscore like it is in the MEDS enum now and not without the underscore.
Right? @dyceron 
image

dyceron:
The enum uses an underscore then I guess so. How would that have happened?

ThanatosGit:
Guess it was a typo by us. You can also see it in the diff from the removed lines that we had "powerbeam" there.
https://github.com/randovania/open-samus-returns-rando/pull/591/changes#diff-99fee7868f49b74b5ea3062af19a61e343850df3d52d0904d72c8c1e81e63fdfL246-L249

dyceron:
Ah I see that now

ThanatosGit:
Can we merge it then? Not blocking the MEDS bump in RDV has become more of a priority.
Can still use new API in upcoming PRs (and at least one fixme actually requires a change in MEDS)

dyceron:
Yes we can merge

dyceron (approved):

PR #594, opened 2026-05-03
Activate / add "reveal map" feature

### LINKED ISSUE #36: Reveal map from the start
Might be found in minimap.lua

ThanatosGit:
- Updated external repo msr-remote-connector
- updated the resulting ips files
- added a delay of 1 second to save the game because the revealed map would not be saved otherwise (the "vanilla" save game is created before the `ScanVisitDiscoverEverything` runs)
"[MAINTAIN]" means "keep whatever was there before".

See also: https://github.com/randovania/msr-remote-connector/pull/5

Fixes #36


dyceron:
What is "sssb" again?

ThanatosGit:
You have to tell the function what are the types of the parameters you use for the delayed function.
SaveGame function takes string, string, string and a Boolean.

dyceron (approved):

PR #605, opened 2026-06-19
Option for "charge_door_burst_buff" and fix for area item collection percent

ThanatosGit:
Should maybe play a full seed before making a release with this but that's for later


dyceron (approved):

ThanatosGit:
Added final boss logic +/- logic

dyceron (approved):

